PR #1438, opened 2015-03-12
Copy the kernel's definition of msgid64_ds verbatim.

khuey:
The discrepancy doesn't affect x86 (32 or 64 bit) but it does break ARM.

### DIFF
--- a/src/kernel_abi.h
+++ b/src/kernel_abi.h
@@ -497,9 +497,18 @@ struct BaseArch : public wordsize, public FcntlConstants {
   struct shmid64_ds {
     ipc64_perm shm_perm;
     size_t shm_segsz;
-    uint64_t shm_atime_only_little_endian;
-    uint64_t shm_dtime_only_little_endian;
-    uint64_t shm_ctime_only_little_endian;
+    __kernel_time_t shm_atime;
+#if __BITS_PER_LONG != 64
+    unsigned long unused1;
+#endif
+    __kernel_time_t shm_dtime;
+#if __BITS_PER_LONG != 64
+    unsigned long unused2;
+#endif
+    __kernel_time_t shm_ctime;
+#if __BITS_PER_LONG != 64
+    unsigned long unused3;
+#endif
     __kernel_pid_t shm_cpid;
     __kernel_pid_t shm_lpid;
     __kernel_ulong_t shm_nattch;


froydnj:
How does it break ARM, out of curiosity?

khuey:
I'm not entirely sure.  I imagine its a structure packing issue.

froydnj:
Huh, weird.  If it was structure packing, I'd expect the size-checking assertions to fire.

khuey:
Yes, that's what it "breaks".

src/kernel_abi.h
```diff
@@ -497,9 +497,18 @@ struct BaseArch : public wordsize, public FcntlConstants {
   struct shmid64_ds {
     ipc64_perm shm_perm;
     size_t shm_segsz;
-    uint64_t shm_atime_only_little_endian;
-    uint64_t shm_dtime_only_little_endian;
-    uint64_t shm_ctime_only_little_endian;
+    __kernel_time_t shm_atime;
+#if __BITS_PER_LONG != 64
+    unsigned long unused1;
+#endif
```
rocallahan: This doesn't look right. `__BITS_PER_LONG` here would be rr's `__BITS_PER_LONG`; to emulate what `asm-generic/shmbuf.h` is doing, we'd need to make this conditional on the `wordsize` template parameter, which I don't think we can do easily.

I think the simplest thing to do is just provide separate definitions per architecture.

khuey: I don't understand.  The issue here is structure packing.  The sizes of the variables themselves are already correct on all three architectures, but on x86-32 a uint64_t can be 32 bit aligned, while on ARM 64 bit values require 8 byte alignment.  Thus declaring it with 64 bit values in rr while the kernel uses two 32 bit values leads to incorrect structure packing.

rocallahan: Right, I understand that. But if we add this patch, then when we build rr 64-bit, `X86Arch::shmid64_ds` will contain
````
    __kernel_time_t shm_atime;
    __kernel_time_t shm_dtime;
    __kernel_time_t shm_ctime;
````
which is incorrect since `X86Arch::__kernel_time_t` is only 32 bits.

khuey: Aha, I had missed that there are separate definitions of __kernel_time_t.  Alright, yes, we'll need to split it up by arch.

PR #1446, opened 2015-03-19
Invoke python more precisely

nsajko:
Specifying the exact version makes python invocation less ambiguous and
eases its usage on systems which don't link python to python2.
Hopefully nowadays the python2 symlink is installed on all modern
distributions.


nsajko:
Thanks :)

PR #1462, opened 2015-04-24
Fix warning on glibc 2.20

garrison:
`_BSD_SOURCE` is deprecated in glibc 2.20, causing a compilation warning. According to the [feature_test_macros man page](http://man7.org/linux/man-pages/man7/feature_test_macros.7.html), the warning can be silenced while still allowing compatibility with older versions of glibc by defining *both* `_BSD_SOURCE` and `_DEFAULT_SOURCE`.


garrison:
Nevermind; withdrawing this.  It is already defined in `Command.h`.

PR #1469, opened 2015-04-29
Fix build dependencies so -jN builds work

luser:
r? @froydnj  maybe? We were missing a couple of `add_dependencies` of targets on `Generated`, and we were also missing mkdirs in a bunch of commands that cp files to the `32` subdirectory. With this change I can clobber and make -jwhatever and it builds fine on my machine. (4 physical cores, 8 virtual cores).


froydnj:
Merged as 3c6f87c660

PR #1528, opened 2015-07-18
Add a `when-ticks` command

glandium:
Similar to the `when` command which displays the current event number,
the `when-ticks` command displays the current ticks count.


glandium:
Note I'm not sure the checked values in the test are right to work anywhere.

PR #1533, opened 2015-07-27
Implement support for the two trivial cases of the personality syscall.

khuey:
This is needed for pulseaudio, which Firefox seems to launch in some circumstances.


froydnj:
Merged as 32848ef.

PR #1551, opened 2015-10-24
add a -d option to replay for specifying a custom debugger

froydnj:
@huonw, can you verify that this actually works with rust-gdb?


huonw:
It seems to work perfectly!

froydnj:
Thanks for checking!

PR #1570, opened 2015-10-30
Fix superfluous mkstemp call in xattr test case.

fracting:
This also fix an access to uninitialized string `path`.


fracting:
Ah, sorry, I mess up the branch without rebasing on latest remote HEAD, I can rebase again once my fnctl patch is committed.

PR #1694, opened 2016-04-25
Add missing <math.h> header

skitt:
Since Scheduler.cc uses maths functions (floor() etc.) it needs to
include math.h.

Signed-off-by: Stephen Kitt <steve@sk2.org>


froydnj:
Thanks, committed as f3adbd32

PR #1831, opened 2016-09-27
PREVENT_SWITCH after the final signal before restarting a syscall.

khuey:
Would be interested to know if this fixes `block_intr_sigchld` for you, @Keno.  I wasn't able to reproduce it myself with `stress`.


Keno:
Seems good to me, I haven't been able to reproduce this failure again after this.

khuey:
Actually that didn't work at all.  But I can reproduce this now, by enabling the no-syscallbuf variant of the test.

khuey:
Note to self: ptrace_seize tests the fatal or SIGSTOP bit of this.

PR #1859, opened 2016-10-21
Testing CI

khuey:
Once more unto the breach.


khuey:
Looks like it's working pretty well.

PR #1909, opened 2016-11-26
Fix incorrect use of recreate_shared_mmap

Keno:
This function is only supposed to be used with mappings that were previously
created using create_shared_mmap. MonitorSharedMemory tried to use it on an
outside MAP_SHARED mapping, which happened to work if you got lucky and
extract_name came up with a name that was valid. However, since the name
in general doesn't have to follow any particular scheme, it was easy to come
up with one that wasn't (e.g. that contained slashes). Add a new function
(steal_mapping) that does what was intended here: replace an arbitrary mapping
with one shared with rr.


Keno:
My 100th rr commit apparently. Dang should have kept it for something less mundane ;)

PR #1911, opened 2016-11-27
Fix -C for mmap entirely beyond its backing file

Keno:
This was problematic both during record and during replay. During record,
we'd get -1 back from read_bytes_fallible since we can't read any bytes
from the mapping, and during replay, we'd run into problems, because
the memory map looks differently due to us creating special SIGBUS
regions. Address both and add a test to checksum_sanity that always
covers this case.


Keno:
Hmm, although this fixes my reduced test case, it does not fix my original one, so I'll have to take another pass

PR #1956, opened 2017-01-31
[DO NOT MERGE] Testing new CI configuration

Keno:


Keno:
Ok. I think this is all set up correctly now. Of course tests don't yet pass, but that's a separate issue(s).

PR #1995, opened 2017-02-26
Rename verify_not_option to improve clarity

danobi:
Renaming `verify_not_option` to `check_if_option` improves clarity
because we get to avoid double negative if-statements.


danobi:
On second thought, this really isn't necessary.

PR #2019, opened 2017-04-17
Fix typo in GdbConnection.cc

ojura:
I was taking a look at my commits from some time ago and discovered this mistake.
We should bail if the connection is not alive anymore, not the other way around.

I am amazed that this typo did not break anything :-)


khuey:
Thanks!

PR #2063, opened 2017-07-20
When building librr.so, tell CMake to set rpath on install

Keno:
This is required when attempting to install an rr built with librr.so,
into a global prefix (executing from the build directory worked find
even without this change).


khuey:
Thanks!

PR #2139, opened 2017-11-21
[Chaosutil] Define linux-only headers under __linux__.

dcci:
This is a nop for Linux builds, but improves portability to other
systems. In particular, on FreeBSD, sys/user is a kernel only header
that causes preprocessor time errors if used in userland.

cc: @rocallahan 
cc: @sfwhittaker
cc: @espindola


dcci:
I and Robert agreed that it's too premature to incrementally submit patches for FreeBSD until we have something functional on FreeBSD. Close this for now.

PR #2147, opened 2017-11-27
[test-monitor] Include headers used explicitly.

dcci:
On FreeBSD, this fails because the file uses std::string and
errno without including the correct header. On Linux, it happens
to work because those headers are transitively included by some
other header. Let's bee more explicit here.

### DIFF
--- a/src/test-monitor/test-monitor.cc
+++ b/src/test-monitor/test-monitor.cc
@@ -1,4 +1,5 @@
 #include <dirent.h>
+#include <errno.h>
 #include <limits.h>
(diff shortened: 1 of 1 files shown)


dcci:
cc: @rocallahan 
cc: @espindola 
cc: @sfwhittaker

dcci:
@rocallahan Robert, the buildbots are failing with a weird error. Can you please take a look?
Thanks!

khuey:
Seems to just be a glitch in the matrix. retriggering the buildbot got it past there.

PR #2415, opened 2019-12-04
Add RR_APPEND_LOG_FILE.

tbosch:
This allows to capture multiple runs of RR into
a single log file.


tbosch:
Note that this is also useful during test execution, as we call RR 2x there, 1x for record and 2x for replay.

khuey:
Thanks!

PR #2420, opened 2019-12-09
Add docs about the perf event ids.

tbosch:
It took me quite a bit to figure out what these special numbers mean.
I.e. these docs might be useful for any future reader.

### DIFF
--- a/src/PerfCounters.cc
+++ b/src/PerfCounters.cc
@@ -110,6 +110,16 @@ struct PmuConfig {
 };
 
 // XXX please only edit this if you really know what you're doing.
+// event = 0x5101c4:
+// - 51 = generic PMU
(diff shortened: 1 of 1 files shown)


khuey:
What were you trying to figure out?

tbosch:
Just where this number `0x5101c4` came from and what it means exactly.

tbosch:
I was trying to reproduce sporadic failures that we saw when using rr ("Got 0 branch events, expected at least 500"). So I was trying to use `perf stat -e 0x5101c4 ...` to reproduce this, but the calls to `perf` and `rr` didn't always fail at the same time.

So I was trying to investigate where the number `0x5101c4` came from...

After some debugging, our issue seems to be https://github.com/mozilla/rr/issues/2422

PR #2423, opened 2019-12-16
Roll forward: "Detect when the kernel multiplexed hw counters."

tbosch:
Even though we now pin counters, tests show that multiplexing
is still happening.

NEW: Fixes the original commit to use tmp_attr properly for the
case of `IN_TXCP`.

This reverts commit 3677e44cf0b996abcbba2546b351126df8f4f03a.

Our use case:
A host is running linux perf now and then to get machine stats, in parallel to `rr`. 
Without this detection, rr silently succeeds.


tbosch:
See discussion on https://github.com/mozilla/rr/pull/2421

PR #2490, opened 2020-04-06
Fix bug in killing all replay tasks

Keno:
Obviously replay tasks need to check against the real
thread group id, not the recorded thread group id.
Sorry about that.


Keno:
Would be good to have a test that catches this. Obviously the task_map iteration order is non-deterministic, but I figure if we just spawn 20 tasks or so and then try to kill them all, that should give us a good chance to reproduce the hang.

PR #2526, opened 2020-04-21
More thoroughly handle exit event race

Keno:
When I originall wrote this code, I had somewhat optimistically
added an assertion that we would only see the race in the case
of external fatal signals, in the exepectation that we would
see stable exit events in time to record proper exit events.
However, if there is a large number of threads waiting to exit,
it's quite possible for the kernel to not deliver us those
notifications until we're already trying to kill the tasks.
In principle it's a simple matter to change the assertion.
However, I also noticed that in this case, we don't write
any exit task events, which I know can confuse external tooling,
so let's refactor this whole thing to have RecordSession have
it's own copy of the killing code (half of it was already
record specific anyway), and properly record exit events
for everything that dies.

Also adds a test that tries to provoke these kinds of races.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -781,6 +781,8 @@ set(BASIC_TESTS
   fork_brk
   fork_child_crash
   fork_many
+  futex_exit_race
+  futex_exit_race_sigsegv
   futex_pi
   futex_priorities
   fxregs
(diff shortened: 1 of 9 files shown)


Keno:
Actually, this doesn't quite work correctly. I have a variation on this test for which this fails. Hang on.

Keno:
Alright, this should work now and I've updated the PR with the test variation.

Issue #1373, opened 2014-11-18
Support ARM


khuey:
Should file this since I'm working on it.

https://github.com/khuey/rr/compare/arm

Requires a couple kernel patches at the moment too.

ignoramous:
Would this mean that rr could be used to debug android applications?

rocallahan:
Eventually yes, but that's a lot of extra work beyond the ARM support Kyle is working on.

petevine:
Any news on ARM support?

rocallahan:
ARM support is not happening in the forseeable future. We discovered a critical technical issue: ARM processors implement atomic operations using a load-linked/store-conditional pair of instructions, and those operations can fail nondeterministically (from rr's point of view; failures depend on cache state and whether a hardware interrupt occurs between the instructions). So we don't have a performance counter that is deterministic enough for rr to use under those conditions.

To fix this, we'd have to modify rr's design philosophy and instrument all ARM code, perhaps using DynamoRio or something like that. That's a lot of work and the cost/benefit for Mozilla doesn't seem to be there right now.

If you really need rr-like functionality for ARM and Android, I recommend buying UndoDB from Undo Software.

matt2909:
That is a strange comment, can you explain what is lacking with an approach such as the linux kernel takes for atomic operations:

http://lxr.free-electrons.com/source/arch/arm/include/asm/atomic.h#L41

rocallahan:
There is no difficulty implementing atomic operations. The problem is that they can disturb the performance counters.

For example, suppose we're using the number of retired instructions, measured via HW performance counters, as our progress counter. Suppose we record a simple program that just does an atomic increment using the code sequence you referenced. Suppose that the LL/SC pair succeeds the first time and we record N instructions executed. Now suppose we replay the execution but this time, a hardware interrupt occurs between the `ldrex` and the `strex` instructions, forcing the `strex` to fail and the code to execute another iteration of the loop. The program completes but performance counters report that we have executed N+4 instructions.

This effect means that performance counters are not 100% reliable for our purposes, which makes rr's zero-instrumentation approach infeasible.

Keno:
I've been reading more about ARM performance counters. It seems that at least the newer ARM chips can count failed strex instructions. I wonder whether setting ticks to `branches taken - failed strex` would be consistent enough for our purposes (It's of course possible for there to be branches in the ll/sc pair, but I don't know how common that is in the real world).

rocallahan:
[...]

PR #2551, opened 2020-05-17
Add generic syscall numbers to syscall table

Keno:
These "generic" syscall numbers are in use by newer architectures
such as aarch64 (ref #1373). They are defined in the following file
in the linux kernel:
https://github.com/torvalds/linux/blob/master/include/uapi/asm-generic/unistd.h

### DIFF
--- a/src/syscalls.py
+++ b/src/syscalls.py
@@ -7,16 +7,17 @@ class BaseSyscall(object):
     """
 
     # Take **kwargs and ignore to make life easier on RegularSyscall.
-    def __init__(self, x86=None, x64=None, **kwargs):
+    def __init__(self, x86=None, x64=None, generic=None, **kwargs):
         assert x86 or x64       # Must exist on one architecture.
         self.x86 = x86
         self.x64 = x64
+        self.generic = generic
         assert len(kwargs) is 0
 
 class RestartSyscall(BaseSyscall):
     """A special class for the restart_syscall syscall."""
-    def __init__(self, x86=None, x64=None):
-        BaseSyscall.__init__(self, x86=x86, x64=x64)
+    def __init__(self, x86=None, x64=None, generic=None):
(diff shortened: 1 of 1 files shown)


khuey:
I'm just going to assume you've copied them all correctly :)

Keno:
:). I did do a second pass through before submitting just to make sure. So far it seems to be working.

src/syscalls.py
```diff
@@ -7,16 +7,17 @@ class BaseSyscall(object):
...
         self.x64 = x64
+        self.generic = generic
         assert len(kwargs) is 0
 
 class RestartSyscall(BaseSyscall):
     """A special class for the restart_syscall syscall."""
-    def __init__(self, x86=None, x64=None):
-        BaseSyscall.__init__(self, x86=x86, x64=x64)
+    def __init__(self, x86=None, x64=None, generic=None):
+        BaseSyscall.__init__(self, x86=x86, x64=x64, generic=None)
```
khuey, line 20: Shouldn't this pass through?

src/syscalls.py
```diff
@@ -25,8 +26,8 @@ class UnsupportedSyscall(BaseSyscall):
     can be displayed in error messages, if nothing else.  They also serve as
     useful documentation.
     """
-    def __init__(self, x86=None, x64=None):
-        BaseSyscall.__init__(self, x86=x86, x64=x64)
+    def __init__(self, x86=None, x64=None, generic=None):
+        BaseSyscall.__init__(self, x86=x86, x64=x64, generic=None)
```
khuey, line 30: And this?

src/syscalls.py
Keno: oops, I triple checked the numbers, but forgot about the logic ;) Will fix

Keno: #2553

PR #2552, opened 2020-05-17
Clean up misc x86 assumptions

Keno:
This collects all the x86 assumptions I encountered while porting rr to aarch64
that are not part of the core rr logic. That said, I disabled a good number
of tests in my porting effort, so several of them certainly still have x86
assumptions or are x86 specific.

### DIFF
--- a/src/AutoRemoteSyscalls.cc
+++ b/src/AutoRemoteSyscalls.cc
@@ -588,7 +588,7 @@ void AutoRemoteSyscalls::finish_direct_mmap(
    * recording. */
   {
     AutoRestoreMem child_str(*this, backing_file_name.c_str());
-    fd = infallible_syscall(syscall_number_for_open(arch()),
+    fd = infallible_syscall(syscall_number_for_openat(arch()), -1,
                             child_str.get().as_int(),
                             backing_file_open_flags);
   }
--- a/src/Task.cc
+++ b/src/Task.cc
@@ -2705,6 +2705,23 @@ static void spawned_child_fatal_error(const ScopedFd& err_fd,
   _exit(1);
 }
 
+static void disable_tsc(int err_fd) {
+  /* Trap to the rr process if a 'rdtsc' instruction is issued.
+   * That allows rr to record the tsc and replay it
+   * deterministically. */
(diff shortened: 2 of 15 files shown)


src/AutoRemoteSyscalls.cc
```diff
@@ -588,7 +588,7 @@ void AutoRemoteSyscalls::finish_direct_mmap(
    * recording. */
   {
     AutoRestoreMem child_str(*this, backing_file_name.c_str());
-    fd = infallible_syscall(syscall_number_for_open(arch()),
+    fd = infallible_syscall(syscall_number_for_openat(arch()), -1,
```
khuey, line 591: I guess passing -1 here works because the path is always absolute?

src/Task.cc
```diff
@@ -2754,12 +2754,14 @@ static void set_up_process(Session& session, const ScopedFd& err_fd,
     setsid();
   }
 
+#if defined(__i386__) || defined(__x86_64)
```
khuey, line 2757: I think I'd prefer you put this in an arch function.

src/test/eventfd.c
```diff
@@ -13,7 +13,9 @@ static void do_test(int fd) {
 }
 
 int main(void) {
+#ifdef SYS_eventfd
   do_test(syscall(SYS_eventfd, 0));
```
khuey, line 17: Seems like it'd be better to just use eventfd2.

Keno:
> Aren't Aarch64 pages also 4KB?

It's configurable

src/AutoRemoteSyscalls.cc
Keno: Yes, this is the intended way to open paths that are guaranteed to be absolute.

src/test/eventfd.c
Keno: The libc wrapper on the next line will test that. I think it's good to just have something that calls the old syscall also, just to make sure it doesn't bitrot.

src/Task.cc
Keno: Ok

src/AutoRemoteSyscalls.cc
khuey: Makes sense. I thought this was AT_FDCWD at first and had to look up its value :)

PR #2561, opened 2020-05-18
Add minimal architecture definition for Aarch64

Keno:
The first commit re-arranges some of our ptrace support to avoid trying to access `Arch::` data structures that don't exist on Aarch64 because these ptrace commands are not support. The remaining two commits add a skeleton implementation of the AArch64 arch definition. `Registers` support is missing, but there's some trickiness there, so that'll be a separate PR.


src/kernel_abi.h
```diff
@@ -1969,6 +1972,85 @@ struct X86Arch : public BaseArch<SupportedArch::x86, WordSize32Defs> {
...
+    blksize_t st_blksize;
+    int __pad2;
+    blkcnt_t st_blocks;
+    struct timespec st_atim;
+    struct timespec st_mtim;
+    struct timespec st_ctim;
+    int __rr_unused[2];
+  };
+
+  struct stat64 {
```
khuey, line 2002: AArch64 doesn't have a stat64 afaict. So why is this here?

src/record_syscall.cc
```diff
@@ -2556,6 +2577,20 @@ static Switchable prepare_ptrace(RecordTask* t,
...
+    case PTRACE_POKEDATA:
+    case PTRACE_PEEKUSER:
+    case PTRACE_POKEUSER:
+    case PTRACE_GETREGS:
+    case PTRACE_GETFPREGS:
+    case PTRACE_GETFPXREGS:
+    case PTRACE_SETREGS:
+    case PTRACE_SETFPREGS:
+    case PTRACE_SETFPXREGS:
+      prepare_ptrace_legacy<Arch>(t, syscall_state);
```
khuey, line 2592: I didn't check very hard that you copy/pasted the code here correctly :)

src/kernel_abi.h
Keno: syscalls.py uses fstatat64 for all architectures. There should probably be a separate entry for fstatat that uses regular struct stat, then this can be dropped. Will try that.

Keno: Actually, even if we do that, it'll still fail to build because the code generated from syscalls.py is not stratified by architecture. Doing that seems a bit overkill. Should we `typedef sts64 stat` instead on architectures that don't have the stat64 syscall?

Keno: Actually a typedef doesn't work because `Arch::stat64` is already the syscall, but we could do
```
struct stat64 : public stat {};
```
for 64 bit archs.

khuey: Ok, yeah, now that I look at it I see x86_64 has the same thing going on. We could clean that up in a followup with the inheritance you suggest but no need to block this.

src/Monkeypatcher.cc
```diff
@@ -930,6 +944,17 @@ void patch_at_preload_init_arch<X64Arch>(RecordTask* t,
                                         params.syscall_patch_hooks);
 }
 
+template <>
+void patch_at_preload_init_arch<AA64Arch>(RecordTask* t,
+                                         Monkeypatcher&) {
+  auto params = t->read_mem(
+      remote_ptr<rrcall_init_preload_params<X64Arch>>(t->regs().arg1()));
```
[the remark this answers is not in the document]
Keno: Sorry, snuck in while splitting this into a sensibly sized commit ;). Fixed.

src/GdbConnection.h
```diff
@@ -581,6 +581,7 @@ class GdbConnection {
   enum {
     CPU_64BIT = 0x1,
```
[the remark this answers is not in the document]
Keno: Done

src/record_syscall.cc
Keno: Well, I didn't touch it, but hopefully the tests would catch it if I had ;). I did catch #2559 while doing this copy pasting so hopefully it's actually better than before :).

src/kernel_abi.h
Keno: I included the inheritance thing here. Seems to work fine.

PR #2562, opened 2020-05-18
Add a record flag to unmap vDSO in new tasks

Keno:
This adds `--unmap-vdso` as a record flag, which will unmap
the vDSO from every task's address space after exec and clear
the corresponding entry from the auxv list.
The vDSO patching machinery is a fair bit of complexity. When
porting to new architectures, it can be useful to first try
without it. I suppose it may also be useful if there is ever
a kernel version that has a vDSO entrypoint the Monkeypatcher
doesn't know about.

I'm hoping this will be useful to the next person trying to port to
a new architectures (cc @vchuravy ;) ).

### DIFF
--- a/src/RecordCommand.cc
+++ b/src/RecordCommand.cc
@@ -100,7 +100,9 @@ RecordCommand RecordCommand::singleton(
     "  --trace-id                 Sets the trace id to the specified id.\n"
     "  --copy-preload-src         Copy preload sources to trace dir\n"
     "  --stap-sdt                 Enables the use of SystemTap statically-\n"
-    "                             defined tracepoints\n");
+    "                             defined tracepoints\n"
+    "  --unmap-vdso               Forcibly unmaps the vdso. This is mostly\n"
+    "                             useful for debugging rr itself.");
 
 struct RecordFlags {
   vector<string> extra_env;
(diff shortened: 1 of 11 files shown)


src/RecordCommand.cc
```diff
@@ -100,7 +100,9 @@ RecordCommand RecordCommand::singleton(
     "  --trace-id                 Sets the trace id to the specified id.\n"
     "  --copy-preload-src         Copy preload sources to trace dir\n"
     "  --stap-sdt                 Enables the use of SystemTap statically-\n"
-    "                             defined tracepoints\n");
+    "                             defined tracepoints\n"
+    "  --unmap-vdso               Forcably unmaps the vdso. This is mostly\n"
+    "                             Useful for debugging rr itself.");
```
khuey, line 105: Forcibly
and a lowercase useful.

Keno:
> If this flag is specified everything that calls AddressSpace::vdso needs to check has_vdso first, no? I don't see changes that do that?

`vdso()` is only used in the MonkeyPatcher, so the checks added here adequately protect it. Which is a good thing too, since the application could always unmap vdso itself first thing.

khuey:
It's used in `find_sycall_instruction`, which I suppose is likely getting called before you unmap the VDSO.

khuey:
Yes. That gets called at the beginning of `process_execve`.

Can you add a comment to `find_syscall_instruction` noting that it runs after exec and before --unmap-vdso or the application itself have a chance to mess with it?

Keno:
Done.

PR #2563, opened 2020-05-18
Aarch64 GDB register mappings and trace format support

Keno:
Add the GDB mappings and import the GDB .xml files so that the GdbServer will work as soon as we hook up register support. Also add a representation of aarch64 in the trace format.

### DIFF
--- a/src/Registers.cc
+++ b/src/Registers.cc
@@ -97,7 +97,6 @@ template <> struct RegisterInfo<rr::X64Arch> {
 
 template <> struct RegisterInfo<rr::ARM64Arch> {
   static bool ignore_undefined_register(GdbRegister) {
-    FATAL() << "Unimplemented";
     return false;
   }
   static const size_t num_registers = DREG_NUM_LINUX_AARCH64;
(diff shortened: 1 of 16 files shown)


src/Registers.cc
```diff
@@ -153,7 +152,24 @@ RegisterInfo<rr::X64Arch>::Table RegisterInfo<rr::X64Arch>::registers = {
   RV_X64(GS_BASE, gs_base),
 };
 
+#define RV_AARCH64(gdb_suffix, name) RV_ARCH(gdb_suffix, name, rr::ARM64Arch, /* empty */)
+#define RV_AARCH64_WITH_MASK(gdb_suffix, name, comparison_mask, size)          \
+  RV_ARCH(gdb_suffix, name, rr::ARM64Arch, COMMA comparison_mask COMMA size)
```
khuey, line 157: Move these up with the other RV_ macros.

Keno:
Good to go?

khuey:
Yes, sorry, Github didn't send me any notification here.

PR #2564, opened 2020-05-18
Add aarch64 registers definition

Keno:
This one's a bit tricky, because the aarch64 syscall ABI differs from x86. Rather than overwriting the syscall number with the return value, instead the kernel overwrites arg1. It does store the value of arg1 at entry internally (for use in restart_syscall), but this is not exposed via ptrace. Also, the currently executing system call is a separate regset, rather than being part of the regular NT_PRSTATUS regset as it is on x86. It's not clear to me whether or not `original_syscall` should be part of `Registers` on aarch64 or not. If it is, `Task` needs to learn how to deal with multiple regsets being potentially independently dirty. The alternative would be to have a `Task::change_syscall` that is eager on aarch64 and modifies the Task's registers on x86. I have a commit that does that also, but thought this might be cleaner. Having implemented it, I'm not sure. Neither feels particularly satisfying.
Similar story with `orig_arg1`. We often need to look at the original arg1 at Task exit both in record and in replay, so we need to store it somewhere, but it's not clear that `Registers` is the right place for it. During record, we generally already do have the original arg1 stored in the syscall state anyway, but we currently don't have a good place to stash it during replay. Anyway, thoughts appreciated.

### DIFF
--- a/src/RecordTask.cc
+++ b/src/RecordTask.cc
@@ -516,7 +516,7 @@ void RecordTask::on_syscall_exit_arch(int syscallno, const Registers& regs) {
 
   switch (syscallno) {
     case Arch::set_robust_list:
-      set_robust_list(regs.arg1(), (size_t)regs.arg2());
+      set_robust_list(regs.orig_arg1(), (size_t)regs.arg2());
       return;
     case Arch::sigaction:
     case Arch::rt_sigaction:
@@ -524,7 +524,7 @@ void RecordTask::on_syscall_exit_arch(int syscallno, const Registers& regs) {
       update_sigaction(regs);
       return;
     case Arch::set_tid_address:
-      set_tid_addr(regs.arg1());
+      set_tid_addr(regs.orig_arg1());
       return;
     case Arch::sigsuspend:
     case Arch::rt_sigsuspend:
@@ -1172,7 +1172,7 @@ void RecordTask::set_siginfo(const siginfo_t& si) {
 
 template <typename Arch>
 void RecordTask::update_sigaction_arch(const Registers& regs) {
-  int sig = regs.arg1_signed();
+  int sig = regs.orig_arg1_signed();
   remote_ptr<typename Arch::kernel_sigaction> new_sigaction = regs.arg2();
   if (0 == regs.syscall_result() && !new_sigaction.is_null()) {
     // A new sighandler was installed.  Update our
--- a/src/Registers.h
+++ b/src/Registers.h
@@ -198,6 +198,7 @@ public:
   }
 
   SYSCALL_REGISTER(syscall_result, eax, rax, x[0]);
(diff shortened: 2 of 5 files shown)


khuey:
It's rather unfortunate that there's no NT_ARM_ORIGX0, though I suppose it's not as bad as the behavior the 32 bit ARM kernel has (or maybe it's been fixed in the intervening five years) where it just outright stomps on r12 and reports garbage to userspace :)

You didn't include code that actually sets orig_x0 so it's not easy to imagine how your manual tracking works. That sounds likely to be complex. Also, what are you doing about the kernel stomping on x7?

Keno:
> You didn't include code that actually sets orig_x0 so it's not easy to imagine how your manual tracking works. That sounds likely to be complex.

I just set it on syscall exit to the value that we recorded for x0 at syscall entry. But you are quite correct, it is a bit annoying - I'm not sure what the best solution is.

> Also, what are you doing about the kernel stomping on x7?

For me it appears to be always setting it to 1, so I just included that in canonicalize_regs, which seems to work well enough. I did mean to look into where in the kernel this happened. Do you have a reference?

Keno:
> It's rather unfortunate that there's no NT_ARM_ORIGX0

If we think that's the best way to do it, I can propose this as a patch upstream. That said, I'm not sure there's ever a situation where we would need to know this, but we didn't observe a syscall or seccomp trap before at which point we could have stored it. If we need it for the syscallbuf, we can always store it in a well known memory location and read it from there.

khuey:
https://elixir.bootlin.com/linux/latest/source/arch/arm64/kernel/ptrace.c#L1814

khuey:
It's rather unfortunate (I'm saying this a lot) that a debugger can't actually inspect the value of x7 at syscall entry on ARM. That said, maybe this doesn't end up affecting us much since the kernel saves/restores it.

It does mean if anyone ever gets an async signal in the middle of a loop where the loop index is in x7 we might be in trouble.

Keno:
> https://elixir.bootlin.com/linux/latest/source/arch/arm64/kernel/ptrace.c#L1814

Huh, interesting. x7 did look correct on syscall entry to me, so I thought it was something strange, but I suspect the reason for this is that syscall entry is reporting during a seccomp stop, so the registers are actually correct, since this only affects syscall stops, which the seccomp stop technically is not.

khuey:
Actually I take that last part back, since it only affects the syscall entry ptrace trap it won't matter for async signals.

Keno:
Yes, I didn't observe any problems with async signals.

khuey:
Alright, I'm convinced this is ok for now.

Keno:
> It's rather unfortunate (I'm saying this a lot) that a debugger can't actually inspect the value of x7 at syscall entry on ARM.

I've sent an email to the mailing list to complain.

PR #2565, opened 2020-05-18
Add rr_page definitions for aarch64

Keno:
Also use `X86Arch::user_desc` throughout rather than relying on it being defined by system headers. This is the type of the thread area, which isn't used on aarch64. That said, there is little harm to still having the field in Task, we just need to make sure it compiles.

### DIFF
--- a/src/AddressSpace.cc
+++ b/src/AddressSpace.cc
@@ -283,6 +283,9 @@ static string find_rr_page_file(Task* t) {
     case x86_64:
       path += "64";
       break;
+    case aarch64:
+      path += "arm64";
+      break;
     default:
       ASSERT(t, false) << "Unknown architecture";
       return path;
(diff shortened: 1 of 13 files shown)


src/AddressSpace.cc
```diff
@@ -373,22 +376,61 @@ static const AddressSpace::SyscallType entry_points[] = {
...
+      val = 3;
+      break;
+    case aarch64:
+      val = 8;
+      break;
+    default:
+      FATAL() << "Syscall stub size not defined for this architecture";
+  }
+  if (arch == NativeArch::arch()) {
+    DEBUG_ASSERT(val == RR_PAGE_SYSCALL_STUB_SIZE);
```
khuey, line 393: What's the benefit of keeping this constant and the assertion?

src/AddressSpace.cc
```diff
@@ -373,22 +376,61 @@ static const AddressSpace::SyscallType entry_points[] = {
...
+      val = 2;
+      break;
+    case aarch64:
+      val = 4;
+      break;
+    default:
+      FATAL() << "Syscall stub size not defined for this architecture";
+  }
+  if (arch == NativeArch::arch()) {
+    DEBUG_ASSERT(val == RR_PAGE_SYSCALL_INSTRUCTION_END);
```
khuey, line 412: Likewise.

src/AddressSpace.cc
Keno: The syscallbuf code makes use of the constant to determine the addresses it needs. The assertion ensures that the constant defined in preload_interface is consistent with the enum values defined here (at least for the current architecture, but it ever gets out of sync, you'd notice before it becomes a stranger problem assuming you're on the problematic architecture).

PR #2568, opened 2020-05-19
Move arch specific ptrace commands into kernel_abi

Keno:
ptrace is highly architecture specific. Different architectures support
different ptrace commands and even when they support the same ones, the
number assigned to the PTRACE_* constants can vary (at least for the
low-numbered commands, the numbering scheme was unified in newer kernels).
If they all had the same numbers across architectures (just with some
architectures missing), we could just put everything in kernel_supplement,
but since that's not the case, kernel_abi is probably the right place for
it. This requries a bit of re-organization, but is fairly mechanical.

While we're here, also take care of arch_prctl, but here, since this
syscall is x86_64 specific, we only need to make sure all values we
use are present in kernel_supplement.h

### DIFF
--- a/src/Task.cc
+++ b/src/Task.cc
@@ -1,9 +1,6 @@
 /* -*- Mode: C++; tab-width: 8; c-basic-offset: 2; indent-tabs-mode: nil; -*- */
 
-#include "Task.h"
-
 #include <asm/prctl.h>
-#include <asm/ptrace.h>
 #include <elf.h>
 #include <errno.h>
 #include <limits.h>
@@ -32,6 +29,8 @@
 #include <sstream>
 
 #include <rr/rr.h>
+
+#include "Task.h"
 
 #include "preload/preload_interface.h"
 
@@ -466,7 +465,7 @@ template <typename Arch>
 static void ptrace_syscall_exit_legacy_arch(Task* t, Task* tracee, const Registers& regs)
 {
   switch ((int)regs.orig_arg1_signed()) {
-    case PTRACE_SETREGS: {
+    case Arch::PTRACE_SETREGS: {
       auto data = t->read_mem(
           remote_ptr<typename Arch::user_regs_struct>(regs.arg4()));
       Registers r = tracee->regs();
@@ -474,7 +473,7 @@ static void ptrace_syscall_exit_legacy_arch(Task* t, Task* tracee, const Registe
       tracee->set_regs(r);
       break;
     }
-    case PTRACE_SETFPREGS: {
+    case Arch::PTRACE_SETFPREGS: {
       auto data = t->read_mem(
           remote_ptr<typename Arch::user_fpregs_struct>(regs.arg4()));
       auto r = tracee->extra_regs();
@@ -482,7 +481,7 @@ static void ptrace_syscall_exit_legacy_arch(Task* t, Task* tracee, const Registe
       tracee->set_extra_regs(r);
       break;
     }
-    case PTRACE_SETFPXREGS: {
+    case Arch::PTRACE_SETFPXREGS: {
       auto data =
           t->read_mem(remote_ptr<X86Arch::user_fpxregs_struct>(regs.arg4()));
       auto r = tracee->extra_regs();
@@ -490,7 +489,7 @@ static void ptrace_syscall_exit_legacy_arch(Task* t, Task* tracee, const Registe
       tracee->set_extra_regs(r);
       break;
     }
-    case PTRACE_POKEUSER: {
+    case Arch::PTRACE_POKEUSR: {
       size_t addr = regs.arg3();
       typename Arch::unsigned_word data = regs.arg4();
       if (addr < sizeof(typename Arch::user_regs_struct)) {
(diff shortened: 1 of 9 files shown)


src/record_syscall.cc
```diff
@@ -2330,7 +2329,12 @@ static Switchable prepare_ptrace(RecordTask* t,
       syscall_state.emulate_result(0);
       break;
     }
-    case PTRACE_OLDSETOPTIONS:
+    case X86Arch::PTRACE_OLDSETOPTIONS:
+      if (t->arch() != x86 && t->arch() != x86_64) {
+        syscall_state.emulate_result(-EINVAL);
```
khuey, line 2334: ptrace returns EIO for invalid requests.

khuey:
I wonder if it would be better if we did what we did for syscalls and gave the PTRACE_BLAH constants that don't exist on a given arch (unique) negative numbers.

Keno:
> The handling of arch_prctl here is weird. non-x86 platforms don't have this syscall at all afaik, so why are you doing piecemeal handing of individual ARCH_BLAH cases instead of just blacklisting the whole syscall and PTRACE_ARCH_PRCTL with architecture assertions?

Fair enough. I was sort of anticipating that it would be added, because it's the new way for applications to map extra copies of the vdso, but you're right, we should just wholesale disable it for the moment.

src/record_syscall.cc
Keno: Ah, right. Good catch, thanks.

Keno:
> I wonder if it would be better if we did what we did for syscalls and gave the PTRACE_BLAH constants that don't exist on a given arch (unique) negative numbers.

That seems like a reasonable idea - let me try that.

Keno:
Updated. See if you like this better.

src/Task.cc
```diff
@@ -721,6 +720,9 @@ void Task::on_syscall_exit_arch(int syscallno, const Registers& regs) {
           break;
         }
         case PTRACE_ARCH_PRCTL: {
```
khuey, line 722: I think you mean Arch::PTRACE_ARCH_PRCTL here.

src/Task.cc
```diff
@@ -721,6 +720,9 @@ void Task::on_syscall_exit_arch(int syscallno, const Registers& regs) {
...
+          if (tracee->arch() != x86_64) {
+            break;
```
khuey, line 724: And then you won't need this.

src/kernel_abi.h
```diff
@@ -1815,6 +1878,20 @@ struct X86Arch : public BaseArch<SupportedArch::x86, WordSize32Defs> {
...
+  static const int PTRACE_GETREGS = Arch64::PTRACE_GETREGS;
+  static const int PTRACE_SETREGS = Arch64::PTRACE_SETREGS;
+  static const int PTRACE_GETFPREGS = Arch64::PTRACE_GETFPREGS;
+  static const int PTRACE_SETFPREGS = Arch64::PTRACE_SETFPREGS;
+  static const int PTRACE_GETFPXREGS = Arch64::PTRACE_GETFPXREGS;
+  static const int PTRACE_SETFPXREGS = Arch64::PTRACE_SETFPXREGS;
+  static const int PTRACE_OLDSETOPTIONS = Arch64::PTRACE_OLDSETOPTIONS;
+  static const int PTRACE_GET_THREAD_AREA = Arch64::PTRACE_GET_THREAD_AREA;
+  static const int PTRACE_SET_THREAD_AREA = Arch64::PTRACE_SET_THREAD_AREA;
+  static const int PTRACE_ARCH_PRCTL = Arch64::PTRACE_ARCH_PRCTL;
```
khuey, line 1891: You can just delete this and let it fall through to the default negative value, since 32 bit x86 doesn't have PTRACE_ARCH_PRCTL.

src/record_syscall.cc
```diff
@@ -3976,6 +3979,7 @@ static Switchable rec_prepare_syscall_arch(RecordTask* t,
       return PREVENT_SWITCH;
 
     case Arch::arch_prctl:
+      DEBUG_ASSERT(t->arch() == x86 || t->arch() == x86_64);
```
khuey, line 3982: Likewise this shouldn't be necessary because the arch_prctl syscallno is -121 on ARM.

src/replay_syscall.cc
```diff
@@ -1126,6 +1126,7 @@ static void rep_process_syscall_arch(ReplayTask* t, ReplayTraceStep* step,
       }
       RR_FALLTHROUGH;
     case Arch::arch_prctl: {
+      DEBUG_ASSERT(t->arch() == x86 || t->arch() == x86_64);
```
khuey, line 1129: ditto

src/Task.cc
Keno: thanks

src/Task.cc
Keno: PTRACE_ARCH_PRCTL can be issues by an x86_64 tracer on an x86 tracee, which this is supposed to catch. I just moved this up from a bit below.

src/kernel_abi.h
Keno: will do

src/record_syscall.cc
Keno: I thought it would be useful to be defensive here, just in case it gets added on another architecture. But I guess such a commit would also add a test that would catch any issue, so I'll get rid of this.

Keno:
Updated again.

src/Task.cc
khuey: Ah good point.

PR #2569, opened 2020-05-20
Miscellaneous aarch64 build fixes

Keno:
Most not particularly interesting. The one thing I thought was kinda interesting is that of the structs we verify, in kernel_abi.h, there's only one that did not match the layout in BaseArch (semid64_ds), so that's fixed also.


src/generate_syscalls.py
```diff
@@ -153,7 +153,7 @@ def write_check_syscall_numbers(f):
     'SyscallEnumsGeneric': lambda f: write_syscall_enum(f, 'generic'),
     'SyscallEnumsForTestsX86': lambda f: write_syscall_enum_for_tests(f, 'x86'),
     'SyscallEnumsForTestsX64': lambda f: write_syscall_enum_for_tests(f, 'x64'),
-    'SyscallEnumsForTestsGeneric': lambda f: write_syscall_enum(f, 'generic'),
+    'SyscallEnumsForTestsGeneric': lambda f: write_syscall_enum_for_tests(f, 'generic'),
```
khuey, line 156: Ah I should have caught that one earlier.

PR #2573, opened 2020-05-22
Aarch64 task running and register logic

Keno:
As discussed in #2564, aarch64 register definition is a bit tricky,
because the kernel does not give us a way to request orig_x0, so we
need to track it ourselves. This means splitting the register format
in the trace from just a default Arch::user_regs_struct to include
the extra orig_x0. Additionally, we need to make sure to copy x0 to
this field on every syscall entry.

### DIFF
--- a/src/AutoRemoteSyscalls.cc
+++ b/src/AutoRemoteSyscalls.cc
@@ -243,6 +243,7 @@ long AutoRemoteSyscalls::syscall_base(int syscallno, Registers& callregs) {
     callregs.set_ip(initial_regs.ip());
   }
 
+  callregs.set_original_syscallno(syscallno);
   callregs.set_syscallno(syscallno);
   t->set_regs(callregs);
 
@@ -254,6 +255,16 @@ long AutoRemoteSyscalls::syscall_base(int syscallno, Registers& callregs) {
       if (t->ip() != callregs.ip()) {
         // We entered the syscall, so stop now
         break;
+      }
+      if (t->ptrace_event() == PTRACE_EVENT_EXIT) {
+        // We died, just let it be
+        break;
+      }
+      if (t->stop_sig() == SIGTRAP && t->get_siginfo().si_code == TRAP_TRACE) {
+        // On aarch64, if we were previously in a syscall-exit stop, continuing
+        // with PTRACE_SINGLESTEP will result in incurring a trap upon execution
+        // of the first instruction in userspace. Ignore such a trap.
+        continue;
       }
       if (ignore_signal(t)) {
         // We were interrupted by a signal before we even entered the syscall
--- a/src/Monkeypatcher.cc
+++ b/src/Monkeypatcher.cc
@@ -937,7 +937,22 @@ void patch_after_exec_arch<X64Arch>(RecordTask* t, Monkeypatcher& patcher) {
 }
 
 template <>
-void patch_after_exec_arch<ARM64Arch>(RecordTask*, Monkeypatcher&) {
+void patch_after_exec_arch<ARM64Arch>(RecordTask* t, Monkeypatcher& patcher) {
+  setup_preload_library_path<ARM64Arch>(t);
+  setup_audit_library_path<ARM64Arch>(t);
+
+  for (const auto& m : t->vm()->maps()) {
+    auto& km = m.map;
+    patcher.patch_after_mmap(t, km.start(), km.size(),
+                             km.file_offset_bytes()/page_size(), -1,
+                             Monkeypatcher::MMAP_EXEC);
+  }
+
+  if (!t->vm()->has_vdso()) {
+    patch_auxv_vdso(t);
+    return;
+  }
+
   FATAL() << "Unimplemented";
 }
 
--- a/src/Registers.h
+++ b/src/Registers.h
@@ -33,6 +33,8 @@ const uintptr_t X86_DF_FLAG = 1 << 10;
 const uintptr_t X86_RF_FLAG = 1 << 16;
 const uintptr_t X86_ID_FLAG = 1 << 21;
 
+const uintptr_t AARCH64_DBG_SPSR_SS = 1 << 21;
+
 /**
  * A Registers object contains values for all general-purpose registers.
  * These must include all registers used to pass syscall parameters and return
@@ -98,6 +100,11 @@ public:
   };
 
   /**
+   * Get the register content to save in the trace.
+   */
+  InternalData get_regs_for_trace() const;
+
+  /**
    * Equivalent to get_ptrace_for_arch(arch()) but doesn't copy.
    */
   InternalData get_ptrace_for_self_arch() const;
@@ -110,6 +117,13 @@ public:
    */
   void set_from_ptrace_for_arch(SupportedArch arch, const void* data,
                                 size_t size);
+
+  /**
+   * Copy from the arch-specific structure returned in get_regs_for_trace()
+   * back into *this
+   */
+  void set_from_trace(SupportedArch arch, const void* data,
+                      size_t size);
 
 #define ARCH_SWITCH_CASE(rettype, x86case, x64case, arm64case)                 \
 (([=](void) -> rettype {                                                       \
@@ -318,8 +332,6 @@ public:
 
   uintptr_t flags() const;
   void set_flags(uintptr_t value);
-  bool singlestep_flag() { return flags() & X86_TF_FLAG; }
-  void clear_singlestep_flag() { set_flags(flags() & ~X86_TF_FLAG); }
   bool df_flag() const { return flags() & X86_DF_FLAG; }
 
   uintptr_t fs_base() const {
@@ -354,11 +366,24 @@ public:
     return u.arm64regs.pstate;
   }
 
+  void set_pstate(uintptr_t pstate) {
+    DEBUG_ASSERT(arch() == aarch64);
+    u.arm64regs.pstate = pstate;
+  }
+
   void set_x7(uintptr_t x7) {
     DEBUG_ASSERT(arch() == aarch64);
     u.arm64regs.x[7] = x7;
   }
   // End of aarch64 specific accessors
+
+  /**
+   * Modify the processor's single step flag. On x86 this is the TF flag in the
+   * eflags register. On aarch64 this is the DBG_SPSR_SS flag in the pstate
+   * register.
+   */
+  bool singlestep_flag();
+  void clear_singlestep_flag();
 
   void print_register_file(FILE* f) const;
   void print_register_file_compact(FILE* f) const;
@@ -477,10 +502,10 @@ private:
           uint64_t pstate;
         };
       };
-      // This is the NT_ARM_SYSTEM_CALL regset
-      int orig_syscall;
       // This is not exposed through GETREGSET. We track it manually
       uint64_t orig_x0;
+      // This is the NT_ARM_SYSTEM_CALL regset
+      int orig_syscall;
     } arm64regs;
   } u;
 };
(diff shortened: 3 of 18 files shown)


src/AutoRemoteSyscalls.cc
```diff
@@ -255,6 +256,16 @@ long AutoRemoteSyscalls::syscall_base(int syscallno, Registers& callregs) {
         // We entered the syscall, so stop now
         break;
       }
+      if (t->ptrace_event() == PTRACE_EVENT_EXIT) {
+        // We died, just let it be
+        break;
+      }
+      if (t->stop_sig() == SIGTRAP && t->get_siginfo().si_code == TRAP_TRACE) {
```
khuey, line 263: If this is AArch64 specific, perhaps it should be ifdefed?

src/AutoRemoteSyscalls.cc
```diff
@@ -255,6 +256,16 @@ long AutoRemoteSyscalls::syscall_base(int syscallno, Registers& callregs) {
         // We entered the syscall, so stop now
         break;
       }
+      if (t->ptrace_event() == PTRACE_EVENT_EXIT) {
+        // We died, just let it be
```
khuey, line 260: Is this a theoretical concern or did you actually see it happen?

src/Monkeypatcher.cc
```diff
@@ -937,7 +937,22 @@ void patch_after_exec_arch<X64Arch>(RecordTask* t, Monkeypatcher& patcher) {
...
+
+  for (const auto& m : t->vm()->maps()) {
+    auto& km = m.map;
+    patcher.patch_after_mmap(t, km.start(), km.size(),
+                             km.file_offset_bytes()/page_size(), -1,
+                             Monkeypatcher::MMAP_EXEC);
+  }
+
+  if (!t->vm()->has_vdso()) {
+    patch_auxv_vdso(t);
```
khuey, line 952: I'm confused, why is this in a !has_vdso branch?

src/Registers.h
```diff
@@ -33,6 +33,8 @@ const uintptr_t X86_DF_FLAG = 1 << 10;
 const uintptr_t X86_RF_FLAG = 1 << 16;
 const uintptr_t X86_ID_FLAG = 1 << 21;
 
+const uintptr_t AARCH64_DBG_SPSR_SS = 1 << 21;
```
khuey, line 36: I'm just assuming this is correct because it's shockingly difficult to find what the layout of the PSTATE is in the ARM Reference Manual.

src/Task.cc
```diff
@@ -1085,54 +1089,62 @@ static bool is_singlestep_resume(ResumeRequest request) {
...
-        // This also takes care of the did_set_breakpoint_after_cpuid workaround case
-      } else if (ti == TrappedInstruction::INT3 &&
-          ip() == address_of_last_execution_resume +
-                       trapped_instruction_len(TrappedInstruction::INT3)) {
-        // INT3 instructions should also be turned into a singlestep here.
+  const siginfo_t& si = get_siginfo();
+  if (arch() == x86 || arch() == x86_64) {
+    uintptr_t status = x86_debug_status();
+    reasons.singlestep = (status & DS_SINGLESTEP) != 0;
+    if (!reasons.singlestep && is_singlestep_resume(how_last_execution_resumed)) {
```
khuey, line 1099: Why did you add the `!reasons.singlestep` condition?

src/Task.cc
```diff
@@ -1234,24 +1251,25 @@ static bool cpu_has_KNL_string_singlestep_bug() {
...
-    LOG(debug) << "Working around KNL single-step hardware bug (cx=" << cx
-               << ")";
-    if (cx > cutoff) {
-      last_resume_orig_cx = cx;
-      Registers r = regs();
-      /* An arbitrary value < cutoff would work fine here, except 1, since
-         the last iteration of the loop behaves differently */
-      r.set_cx(cutoff);
-      set_regs(r);
+  if (cpu_has_KNL_string_singlestep_bug()) {
```
khuey, line 1256: I think it'd be easier just to `if (!cpu_has_KNL_string_singlestep_bug()) { return; }` at the beginning and leave the rest of this function undisturbed.

src/Task.cc
```diff
@@ -1345,13 +1366,35 @@ void Task::flush_regs() {
...
+  if (orig_syscallno_dirty) {
+    uintptr_t syscall = registers.original_syscallno();
+    struct iovec vec = { &syscall,
+                          sizeof(syscall) };
+    LOG(debug) << "Chaning syscall to " << syscall;
+    if (ptrace_if_alive(PTRACE_SETREGSET, NT_ARM_SYSTEM_CALL, &vec)) {
+      orig_syscallno_dirty = false;
     }
   }
+#endif
```
khuey, line 1397:
```
#elif defined(__i386__) || defined(__x86_64__)
  ASSERT(t, !orig_syscallno_dirty);
#endif
```

src/Task.cc
```diff
@@ -1345,13 +1366,35 @@ void Task::flush_regs() {
   if (registers_dirty) {
     LOG(debug) << "Flushing registers for tid " << tid << " " << registers;
     auto ptrace_regs = registers.get_ptrace();
+#if defined(__i386__) || defined(__x86_64)
```
khuey, line 1369: It seems your `__x86_64__` got its right arm torn off in some sort of terrible industrial accident.

src/Task.h
```diff
@@ -454,20 +454,27 @@ class Task {
...
   /**
-   * Set the debug status (DR6 on x86).
+   * Set the debug status (DR6 on x86). Noop on non-x86 architectures.
    */
-  void set_debug_status(uintptr_t status);
+  void set_x86_debug_status(uintptr_t status);
 
   /**
-   * Determine why a SIGTRAP occurred. Uses debug_status() but doesn't
+   * Determine why a SIGTRAP occurred. Ont x86, uses x86_debug_status() but doesn't
```
khuey, line 467: On

src/Task.h
```diff
@@ -1101,6 +1108,9 @@ class Task {
   // True when 'registers' has changes that haven't been flushed back to the
   // task yet.
   bool registers_dirty;
+  // True when changes to the original syscallno in 'registers' have not been
+  // flushed back to the task yet.
```
khuey, line 1112: Add "Some architectures (e.g. AArch64) require a separate ptrace call for this."

src/util.h
```diff
@@ -341,7 +341,7 @@ inline bool is_kernel_trap(int si_code) {
   /* XXX unable to find docs on which of these "should" be
    * right.  The SI_KERNEL code is seen in the int3 test, so we
    * at least need to handle that. */
-  return si_code == TRAP_BRKPT || si_code == SI_KERNEL;
+  return si_code == TRAP_TRACE || si_code == TRAP_BRKPT || si_code == SI_KERNEL;
```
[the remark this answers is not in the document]
Keno: Yes, TRAP_TRACE is single-stepping.

src/Monkeypatcher.cc
Keno: this patches out the vdso address from auxv, so that the application doesn't see it. Perhaps the function needs a better name.

src/Registers.h
Keno: > it's shockingly difficult to find what the layout of the PSTATE is in the ARM Reference Manual.

right? And yes, I think it's correct.

src/AutoRemoteSyscalls.cc
Keno: I saw it happen, but only due to an unrelated bug that caused the tracee to segfault on the first instruction out of a clone taking down the process before rr was done doing its setup. It's not aarch64 specific.

src/remote_code_ptr.h
```diff
@@ -50,6 +50,18 @@ class remote_code_ptr {
   remote_code_ptr decrement_by_bkpt_insn_length(SupportedArch arch) const {
     return remote_code_ptr(ptr - rr::bkpt_instruction_length(arch));
   }
+  remote_code_ptr undo_executed_bkpt(SupportedArch arch) {
+    switch (arch) {
+      case x86:
+      case x86_64:
+        return decrement_by_bkpt_insn_length(arch);
```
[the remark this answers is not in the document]
Keno: Sure.

src/Monkeypatcher.cc
khuey: Lets file an issue on renaming it to `patch_auxv_to_remove_vdso` or something.

src/Task.cc
Keno: The only thing that could happen in any of these branches is setting singlestep to `true`, so it isn't necessary to try any of these things if the kernel already told us it was a single step. I originally had the condition inverted to avoid going through some x86 code paths on aarch64, but then I decided to do it this way.

src/Task.cc
Keno: Indeed. No idea why this ever passed tests

src/Task.cc
Keno: We can't really do that, right, because the ptrace above is allowed to fail?
At most we could do:
```
#elif defined(__i386__) || defined(__x86_64__)
  ASSERT(t, registers_dirty == orig_syscallno_dirty);
#endif
```

src/AutoRemoteSyscalls.cc
Keno: I think powerpc has the same behavior, so it might just be x86 that's the odd one out. I don't think It'll do any harm. That said, I am actually a bit confused why TRAP_TRACE doesn't seem to show up for us on x86, because it seems like it should:
```
static inline int get_si_code(unsigned long condition)
{
	if (condition & DR_STEP)
		return TRAP_TRACE;
	else if (condition & (DR_TRAP0|DR_TRAP1|DR_TRAP2|DR_TRAP3))
		return TRAP_HWBKPT;
	else
		return TRAP_BRKPT;
}
[snip]
	si_code = get_si_code(tsk->thread.debugreg6);
	if (tsk->thread.debugreg6 & (DR_STEP | DR_TRAP_BITS) || user_icebp)
		send_sigtrap(regs, error_code, si_code);
```
I do know why we don't see it here, which is that aarch64 reports an extra single step trap after exiting the syscall, but not sure why we don't see it in general (or maybe we do, and I just never looked closely enough).

src/Task.cc
khuey: Ah, yes, good point.

src/Task.h
Keno: Thanks

src/Task.h
Keno: Will do.

src/util.h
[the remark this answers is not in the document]
Keno: See my comment below. I don't actually know why it's not consistent.

Keno:
Updated.

Keno:
Hold on on merging, there was an intermittent test failure that came up while I tested it that looks like it may have been caused by this. One sec while I look into it.

src/Task.cc
Keno: Actually, that's not quite right either, because it is possible for the registers to be dirty, but orig_syscallno not to be dirty. I think the most we can do is maybe something like:
```
  ASSERT(this, !orig_syscallno_dirty || registers_dirty);
```

khuey: maybe in the x86 `if (registers_dirty)` block add `else { ASSERT(this, !orig_syscallno_dirty); }` then?

Keno: yep

Keno:
Alright, fixed with the change to the assertion as discussed.

khuey:
@Keno the intermittent you mentioned was that assertion?

Keno:
Yes

PR #2577, opened 2020-05-22
Aarch64 build system support

Keno:
Add build flags for aarch64, add the ability to mark tests x86 only
by moving them into the x86/ directory, where they will be built on
x86, but ignored on aarch64 (or other future architectures).


CMakeLists.txt
```diff
@@ -50,7 +50,17 @@ configure_file(
   ${CMAKE_BINARY_DIR}/git_revision.h
 )
 
-set(FLAGS_COMMON "-msse2 -D__MMX__ -D__SSE__ -D__SSE2__ -D__USE_LARGEFILE64 -pthread")
+set(FLAGS_COMMON "-D__USE_LARGEFILE64 -pthread")
+set(supports32bit true)
+set(x86ish false)
```
khuey, line 55: It seems kind of weird to have these start in opposite positions and then have one branch of the if flip each. Maybe just start with the x86 values and have the aarch64 branch clip both.

Keno: The idea here is that eventually aarch64 may support the 32 bit variant also, so true is a better default, but neither it nor any future additions will ever be x86ish

khuey: Alright fair.

PR #2578, opened 2020-05-22
Aarch64 vdso patching support

Keno:
Mostly straightforward. The only difficulty is that the immediate is not at a byte boundary in the instruction on Aarch64 so our current AssemblyTemplate magic wasn't quite powerful enough to handle it.

### DIFF
--- a/src/Monkeypatcher.cc
+++ b/src/Monkeypatcher.cc
@@ -953,7 +953,48 @@ void patch_after_exec_arch<ARM64Arch>(RecordTask* t, Monkeypatcher& patcher) {
     return;
   }
 
-  FATAL() << "Unimplemented";
+  auto vdso_start = t->vm()->vdso().start();
+
+  VdsoReader reader(t);
+  auto syms = reader.read_symbols(".dynsym", ".dynstr");
+
+  static const named_syscall syscalls_to_monkeypatch[] = {
+#define S(n) { "__kernel_" #n, ARM64Arch::n }
+    S(rt_sigreturn), S(gettimeofday), S(clock_gettime), S(clock_getres),
+#undef S
+  };
+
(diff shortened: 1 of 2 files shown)


src/Monkeypatcher.cc
```diff
@@ -953,7 +953,48 @@ void patch_after_exec_arch<ARM64Arch>(RecordTask* t, Monkeypatcher& patcher) {
...
 
-  FATAL() << "Unimplemented";
+  auto vdso_start = t->vm()->vdso().start();
+
+  VdsoReader reader(t);
+  auto syms = reader.read_symbols(".dynsym", ".dynstr");
+
+  static const named_syscall syscalls_to_monkeypatch[] = {
+#define S(n) { "__kernel_" #n, ARM64Arch::n }
+    S(rt_sigreturn), S(gettimeofday), S(clock_gettime), S(clock_getres),
```
khuey, line 963: I think ideally this would share more code with the x86 implementation but I'm not super concerned about it.

src/Monkeypatcher.cc
```diff
@@ -953,7 +953,48 @@ void patch_after_exec_arch<ARM64Arch>(RecordTask* t, Monkeypatcher& patcher) {
...
+        // "mov $syscall_number,x8".
+        patcher.patched_vdso_syscalls.insert(
+            remote_code_ptr(absolute_address + 4));
+        LOG(debug) << "monkeypatched " << syscall.name << " to syscall "
+                    << syscall.syscall_number << " at "
+                    << HEX(absolute_address) << " (" << HEX(file_offset)
+                    << ")";
+      }
+    }
+  }
```
khuey, line 997: Does this not need to `obliterate_debug_info`?

Keno: Ah, it probably should.

PR #2580, opened 2020-05-23
Fix breakpoint reached checks

Keno:
There's legitimate reasons to use the increment_by_bkpt_insn_length
function. However, to check whether a breakpoint was reached, we need
to use the undo_executed_bkpt function instead, since the behavior is
arch specific.


khuey:
What about the one in `AddressSpace::is_breakpoint_in_private_read_only_memory`?

Keno:
I believe that is the one correct use, because it's used to check whether the entirety of the instruction is in the given map.

khuey:
yeah ok.

PR #2581, opened 2020-05-23
Aarch64 grabbag

Keno:
No real unifying theme here other than "I tried things and this is what broke".
Should all be relatively straightforward.

### DIFF
--- a/src/ReplaySession.cc
+++ b/src/ReplaySession.cc
@@ -873,10 +873,6 @@ Completion ReplaySession::emulate_async_signal(
          * check again if we're at the
          * target. */
         LOG(debug) << "    trap was for target $ip";
-        /* (The breakpoint would have trapped
-         * at the $ip one byte beyond the
-         * target.) */
-        DEBUG_ASSERT(!at_target);
 
         pending_SIGTRAP = false;
         t->move_ip_before_breakpoint();
--- a/src/kernel_abi.h
+++ b/src/kernel_abi.h
@@ -2125,6 +2125,26 @@ struct ARM64Arch : public GenericArch<SupportedArch::aarch64, WordSize64Defs> {
   };
   typedef struct user_fpsimd_state user_fpregs_struct;
 
+  struct sigcontext {
+    __u64 fault_addr;
+    user_pt_regs regs;
+    // ISA extension state follows here
+  };
+
+  struct ucontext {
+    unsigned long	uc_flags;
+    ptr<ucontext> uc_link;
+    stack_t		  uc_stack;
+    sigset_t	  uc_sigmask;
+    uint8_t __unused[1024 / 8 - sizeof(sigset_t)];
+    struct sigcontext uc_mcontext;
+  };
+
+  struct rt_sigframe {
+    siginfo_t info;
+    struct ucontext uc;
+  };
+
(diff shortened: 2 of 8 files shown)


src/ReplaySession.cc
```diff
@@ -873,10 +873,14 @@ Completion ReplaySession::emulate_async_signal(
...
-         * at the $ip one byte beyond the
-         * target.) */
-        DEBUG_ASSERT(!at_target);
+        if (t->arch() == x86 || t->arch() == x86_64) {
+          /* (The breakpoint would have trapped
+          * at the $ip one byte beyond the
+          * target.) */
+          DEBUG_ASSERT(!at_target);
+        } else {
+          DEBUG_ASSERT(regs.ip() == t->ip());
```
khuey, line 882: This is just a duplicate of the preceding assertion with undo_executed_bkpt in it.

src/kernel_abi.h
```diff
@@ -2125,6 +2125,26 @@ struct ARM64Arch : public GenericArch<SupportedArch::aarch64, WordSize64Defs> {
   };
   typedef struct user_fpsimd_state user_fpregs_struct;
 
+  struct sigcontext {
+    __u64 fault_addr;
+    user_pt_regs regs;
+    // ISA extension state follows here
+  };
+
+  struct ucontext {
```
khuey, line 2134: Hmm why aren't you RR_VERIFY_TYPE_ARCHing any of these?

src/AddressSpace.cc
```diff
@@ -1817,23 +1817,25 @@ bool AddressSpace::has_exec_watchpoint_fired(remote_code_ptr addr) {
 }
 
 bool AddressSpace::allocate_watchpoints() {
```
khuey, line 1819: Have you given any thought to how this will work on AArch64? I would think you could shoehorn most of the arch specific stuff into Task::set_debug_regs and this code could be largely unchanged.

Keno: No, I haven't looked into watchpoints yet.

src/kernel_abi.h
Keno: rt_sigframe isn't in the uapi. struct sigcontext as defined in the kernel header is useless, because they just stuck a page size u8 array at the end, so taking sizeof of it doesn't give the correct size that the kernel actually stores.

src/ReplaySession.cc
Keno: True under the assumption that all non-x86 architectures have the same breakpoint behavior as aarch64. I wanted to defend against the possibility that that might not be the case such that a future architecture implementer could update the list of platforms here properly. Alternatively, we could just get rid of both branches of this if/else, since I don't think the x86 side adds too much either.

khuey: I'd be fine with removing it entirely.

Keno: Will do

src/AddressSpace.cc
Keno: If you'd prefer I can drop this part of it for now until we figure out how to do watchpoints properly.

khuey: I think that would be good.

Keno:
Dropped the watchpoint `if/else` and replaced the commit adding the assertion in ReplaySession.cc by one that drops the old condition instead.

PR #2584, opened 2020-05-23
More robust exits from coredumping signals

Keno:
Coredumping signals are extremely annoying. Every task in the
same address space, save for the one doing the coredumping
needs to reach its exit event before that task can proceed.
However, we must make sure not to accidentally continue a
task past its exit state, because otherwise it too will hang
until all other tasks have continued. Unfortunately, we don't
have a "ptrace-continue-unless-in-exit-state" command, so we
try to waitpid beforehand to make sure we're not in this
situation. That seems to work reasonably well on x86, but
for some reason not on aarch64, where I'm routinely observing
the waitpid not reporting hat the task has stopped, but
the subsequent ptrace continue, moving it right past the stop.
This change tries to make that race detection slightly less
necessary by not provoking it during coredumping signals
that we inject. Instead, we mark that we're expecting a
ptrace exit event for every task that's about to die and
then refuse to schedule these tasks even if they otherwise
look runnable. This may feel like the return of unstable_exit,
but note that once the exit event arrives, we still process
the exits in an orderly manner, recording each tasks robust
futex changes etc. in order, so we don't lose any of the
advantages that removing unstable exit gave us. Instead,
we're simply avoiding the above described race, and relying
on the existing logic to handle everything properly.

### DIFF
--- a/src/RecordSession.cc
+++ b/src/RecordSession.cc
@@ -1593,6 +1593,17 @@ bool RecordSession::signal_state_changed(RecordTask* t, StepState* step_state) {
          */
         t->pop_signal_delivery();
       }
+
+      // Mark each task in this address space as expecting a ptrace exit
+      // to avoid causing any ptrace_exit reaces.
+      if (is_fatal && is_coredumping_signal(sig)) {
+        for (Task *ot : t->vm()->task_set()) {
+          if (t != ot) {
+            ((RecordTask *)ot)->waiting_for_ptrace_exit = true;
+          }
+        }
+      }
+
       last_task_switchable = can_switch;
       step_state->continue_type = DONT_CONTINUE;
       break;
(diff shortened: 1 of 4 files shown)


Keno:
That said, I'm wondering if upstream would be amenable to an interface that closes this race. I'm thinking setting an exit_state flag just before reaching the PTRACE_EXIT_EVENT and then waitpid can check that and return ECHILD if we've already gone past the PTRACE_EXIT_EVENT. That would close one side. To close the other, add a ptrace option that requires the ptrace to pass SIGKILL as the continue signal to advance out of a PTRACE_EXIT_EVENT. Since it would never ordinarily do that, since SIGKILLs don't get signal stops, it can be sure not to accidentally continue a PTRACE_EXIT_EVENT when the tracer thought some other stop was being processed.

PR #2585, opened 2020-05-24
Misc aarch64 changes (part 4)

Keno:
As before, these are the collection of misc standalone changes that I'm cherry-picking from my WIP branch. Generally they each correspond to one test. I have a couple bigger changes in the oven, but they're still cooking, since I'm still finding ways in which they're not quite right. These changes on the other hand are hopefully small/simple enough that I won't find any surprises here later.

### DIFF
--- a/src/RecordTask.cc
+++ b/src/RecordTask.cc
@@ -1071,6 +1071,7 @@ void RecordTask::emulate_SIGCONT() {
   for (Task* t : thread_group()->task_set()) {
     auto rt = static_cast<RecordTask*>(t);
     LOG(debug) << "setting " << tid << " to NOT_STOPPED due to SIGCONT";
+    rt->clear_stashed_group_stop();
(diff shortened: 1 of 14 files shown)


src/RecordTask.cc
```diff
@@ -1071,6 +1071,7 @@ void RecordTask::emulate_SIGCONT() {
   for (Task* t : thread_group()->task_set()) {
     auto rt = static_cast<RecordTask*>(t);
     LOG(debug) << "setting " << tid << " to NOT_STOPPED due to SIGCONT";
+    rt->clear_stashed_group_stop();
```
khuey, line 1074: Would be nice if this had a test.

Keno: It kinda does. exec_stopsig tests this, but it appears to depend on how quick the signal gets delivered.

PR #2587, opened 2020-05-26
Port ptrace and seccomp tests to aarch64

Keno:
Fairly straightforward, mostly in the tests, but with a small fix to seccomp support and adding support of NT_ARM_SYSTEM_CALL. Very x86-specific tests got moved into `x86/`. There's two tests that test setting hardware breakpoints using the debug registers. That'll need to be adjusted, but since I didn't add watchpoint support to rr proper on aarch64 yet, that'll get done as part of that.


src/RecordSession.cc
```diff
@@ -148,7 +156,7 @@ static bool handle_ptrace_exit_event(RecordTask* t) {
       // that state replay of the SCHED will fail. So detect that state and fix
       // it up.
       if (t->regs().original_syscallno() >= 0 &&
-          t->regs().syscall_result_signed() == -ENOSYS) {
+          looks_like_syscall_entry(t)) {
```
khuey, line 159: I don't really get this. If it's ok for this to always be true on AArch64, why do we need the -ENOSYS check on x86?

src/RecordSession.cc
```diff
@@ -984,7 +992,6 @@ void RecordSession::syscall_state_changed(RecordTask* t,
         return;
       }
       last_task_switchable = PREVENT_SWITCH;
-      t->apply_syscall_entry_regs();
```
khuey, line 987: Uh where did this go?

src/test/seccomp_sigsys_sigtrap.c
```diff
@@ -21,7 +23,7 @@ static void install_filter(void) {
     BPF_STMT(BPF_LD | BPF_W | BPF_ABS, offsetof(struct seccomp_data, nr)),
     /* Jump forward 1 instruction if system call number
        is not SYS_open */
```
khuey, line 25: Change the comment too please.

src/record_syscall.cc
```diff
@@ -1997,6 +1997,7 @@ static void ptrace_verify_set_reg_set(RecordTask* t, size_t min_size,
   auto iov = t->read_mem(remote_ptr<typename Arch::iovec>(t->regs().arg4()));
   if (iov.iov_len < min_size) {
     syscall_state.emulate_result(-EIO);
+    return;
```
khuey, line 2000: Yikes.

src/RecordSession.cc
Keno: I'm not really sure. As the comment immediately below indicates, the check is unreliable anyway. The most common reason we hit this path is due to SECCOMP_RET_KILL, though I think we should probably just record the syscall entry event there rather than relying on this. Thinking a bit more along these lines, I'm wondering whether we should just check if the currently instruction is a syscall instruction and if so always putting a sched event right before the syscall instruction. It seems a bit odd here to hallucinate a syscall entry if the registers happen to look like it.

src/RecordSession.cc
Keno: It was incorrect, because this is a ptrace resume, so it already happened on the original entry, and the ptracer may have modified it.

src/record_syscall.cc
Keno: Yeah, we don't test the error paths at all in the ptrace emulation. We have many issues here beyond this one.

src/test/seccomp_sigsys_sigtrap.c
Keno: Good catch

Keno:
> One horribly terrible question is whether we need to emulate the kernel's squashing of x7 for a ptracer.

I wasn't planning on it, though I don't think it would be particularly hard to do either. My plan is just to leave it alone unless somebody really complains. What we should do though is to fix ptrace recording if the target process is not part of the process tree, so somebody can go record strace with an external pid if necessary and see the broken kernel behavior.

src/RecordSession.cc
Keno: Actually, the original_syscallno test doesn't work either on aarch64, because that'll get reset to -1 before we ever get there (the only place it does work is from SECCOMP_RET_KILL), not that we actually ever asked the tracee about it. I think what we need to do here is:
1. Record ip_at_last_recorded_syscall_exit in addition to ticks_at_last_recorded_syscall_exit.
2. Look at the instruction that's currently executing. If it's an svc, we set the syscall entry, otherwise a sched.

PR #2590, opened 2020-05-26
Work around aarch64 x7 woes

Keno:
During a syscall stop, the kernel lies about x7 and ignores writes to
it. That's usually not a problem as long as we don't look at it. However,
it becomes a problem if we do any AutoRemoteSyscall'ing during the exit
processing. AutoRemoteSyscalls might leave us in either a singlestep
or a syscall exit trap. If the former, then trying to re-apply the
registers that it thought the task had before the AutoRemoteSyscall
will actually end up modifying the real x7 register (because we're
no longer in a syscall stop where the kernel would ignore such a
write).

There's also a few other corner cases (e.g. during restart where
similar tricks need to apply).

In general, we rely on x7 being accurate during syscall entry
(since we expect to enter using a seccomp trap), and then
just copy it to the output event.

This is all a huge mess and I have an ongoing thread with
the kernel maintainers pleading for an option to give some sanity
here. That said, this seems to work in the meantime.

### DIFF
--- a/src/RecordSession.cc
+++ b/src/RecordSession.cc
@@ -235,6 +235,14 @@ static void note_entering_syscall(RecordTask* t) {
     // registers look like they did at the original syscall entry
     // again.
     t->ev().Syscall().regs.set_arg1(t->ev().Syscall().regs.orig_arg1());
+    if (t->arch() == aarch64) {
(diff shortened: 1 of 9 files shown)


Keno:
Oh for sure - I don't like it either. I'll push hard for fixing this upstream.

PR #2591, opened 2020-05-26
Port various tests to aarch64

Keno:
With this and the ptrace PR, I'm seeing all BASIC_TESTS pass on aarch64 (without syscallbuf for the moment, modulo the two ptrace tests I mentioned in the other PR).


khuey:
What does `asm("" : : : "memory");` do on ARM that's special?

Keno:
Nothing, it prevents the compiler from removing the loop (infinite loops are undefined behavior) and breaking the test. Happens on x86 also e.g. when compiling with clang at higher optimization levels.

khuey:
mmm undefined behavior

PR #2593, opened 2020-05-27
Add watchpoint support for aarch64

Keno:
There's a few notable differences (aarch64 watchpoints fire before applying instruction effects, you get the matched address, but not which one fired), but in general the capabilities provided are similar to x86, so the number of changes is reasonable.

### DIFF
--- a/src/AddressSpace.h
+++ b/src/AddressSpace.h
@@ -213,6 +213,17 @@ enum WatchType {
   WATCH_READWRITE = 0x03
 };
 
+enum ArmWatchType {
+  ARM_WATCH_EXEC = 0x0,
+  ARM_WATCH_READ = 0x1,
+  ARM_WATCH_WRITE = 0x2,
+  ARM_WATCH_READWRITE = ARM_WATCH_READ | ARM_WATCH_WRITE
+};
+
+enum ArmPrivLevel {
+  ARM_PRIV_EL0 = 0x2
+};
+
 enum DebugStatus {
   DS_WATCHPOINT_ANY = 0xf,
   DS_SINGLESTEP = 1 << 14,
@@ -582,6 +593,7 @@ public:
    * are known to not be set on singlestep).
    */
   bool notify_watchpoint_fired(uintptr_t debug_status,
+      remote_ptr<void> hit_addr,
       remote_code_ptr address_of_singlestep_start);
   /**
    * Return true if any watchpoint has fired. Will keep returning true until
--- a/src/Task.cc
+++ b/src/Task.cc
@@ -503,7 +503,7 @@ static void ptrace_syscall_exit_legacy_arch(Task* t, Task* tracee, const Registe
         size_t regno =
             (addr - offsetof(typename Arch::user, u_debugreg[0])) /
             sizeof(data);
-        tracee->set_debug_reg(regno, data);
+        tracee->set_x86_debug_reg(regno, data);
       }
       break;
     }
@@ -1062,7 +1062,7 @@ uintptr_t Task::get_debug_reg(size_t regno) {
   return result;
 }
 
-bool Task::set_debug_reg(size_t regno, uintptr_t value) {
+bool Task::set_x86_debug_reg(size_t regno, uintptr_t value) {
   errno = 0;
   fallible_ptrace(PTRACE_POKEUSER, dr_user_word_offset(regno), (void*)value);
   return errno == ESRCH || errno == 0;
@@ -1078,7 +1078,7 @@ uintptr_t Task::get_debug_reg(size_t) {
   return 0;
 }
 
-bool Task::set_debug_reg(size_t, uintptr_t) {
+bool Task::set_x86_debug_reg(size_t, uintptr_t) {
   FATAL_X86_ONLY();
   return false;
 }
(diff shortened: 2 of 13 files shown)


src/Task.cc
```diff
@@ -1534,18 +1565,100 @@ bool Task::set_debug_regs(const DebugRegs& regs) {
...
+      case WATCH_WRITE:
+        ctrl.type = ARM_WATCH_WRITE;
+        break;
+      case WATCH_READWRITE:
+        ctrl.type = ARM_WATCH_READWRITE;
+        break;
+    }
+    ctrl.enabled = 1;
+    ctrl.priv = ARM_PRIV_EL0;
+	  uintptr_t off = (uintptr_t)reg.addr.as_int() % 8;
```
khuey, line 1659: 1. What happened to the indentation here?
2. Does gdb automatically break things that cross the alignment boundary here for us? i.e. it's not going to give us 0xABCD4 with a watchpoint length of 8?

src/AddressSpace.h
```diff
@@ -213,6 +213,17 @@ enum WatchType {
...
 
+enum ArmWatchType {
+  ARM_WATCH_EXEC = 0x0,
+  ARM_WATCH_READ = 0x1,
+  ARM_WATCH_WRITE = 0x2,
+  ARM_WATCH_READWRITE = ARM_WATCH_READ | ARM_WATCH_WRITE
+};
+
+enum ArmPrivLevel {
+  ARM_PRIV_EL0 = 0x2
```
khuey, line 224: It drives me nuts that the numbering of ARM privilege levels is reversed from x86.

src/Task.cc
Keno: Re indentation, not sure. Hmm, not sure. I may have accidentally introduced some tabs or something by editing with vim.

As for the alignment thing, good question. I meant to check that, but forgot.

src/AddressSpace.h
Keno: Yeah, it's tripped me up more than once. Makes sense though. Avoids the negative ring numbers x86 has had to add.

src/Task.cc
khuey: I'm holding off on merging this solely because of this. No rush, but if you're waiting on me I wanted to let you know why this is sitting here.

Keno: Nope fair, just haven't gotten around to it yet.

Keno: Looks like the answer is no, it does not split. I'll adjust the code to do that here.

Keno: Actually looks like I was wrong. It does split, but for some reason the watchpoint doesn't get hit. Let me investigate.

Keno: Turns out I was just testing wrong. This is fine as is, but I'll push the commit to split the watchpoints anyway, since it's small and should provide some extra assurance for any future changes in this GDB behavior.

src/Task.cc
```diff
@@ -1295,21 +1326,34 @@ void Task::resume_execution(ResumeRequest how, WaitRequest wait_how,
...
+        // In KVM virtual machines (and maybe others), singlestepping over CPUID
+        // executes the following instruction as well. Work around that.
+        did_set_breakpoint_after_cpuid =
+          vm()->add_breakpoint(ip() + trapped_instruction_len(singlestepping_instruction), BKPT_INTERNAL);
+      }
+    } else if (arch() == aarch64 && is_singlestep_resume(how_last_execution_resumed)) {
+      // On aarch64, if the last execution was any sort of single step, then
+      // resuming again with PTRACE_(SYSEMU_)SINGLESTEP will cause a debug fault
+      // upon exiting from the current instruction. To prevent this, we must
+      // reset the PSATE.SS bit. (If the last resume was not a single step,
```
khuey, line 1345: PSTATE

src/Task.cc
```diff
@@ -1295,21 +1326,34 @@ void Task::resume_execution(ResumeRequest how, WaitRequest wait_how,
...
+      if (singlestepping_instruction == TrappedInstruction::CPUID) {
+        // In KVM virtual machines (and maybe others), singlestepping over CPUID
+        // executes the following instruction as well. Work around that.
+        did_set_breakpoint_after_cpuid =
+          vm()->add_breakpoint(ip() + trapped_instruction_len(singlestepping_instruction), BKPT_INTERNAL);
+      }
+    } else if (arch() == aarch64 && is_singlestep_resume(how_last_execution_resumed)) {
+      // On aarch64, if the last execution was any sort of single step, then
+      // resuming again with PTRACE_(SYSEMU_)SINGLESTEP will cause a debug fault
+      // upon exiting from the current instruction. To prevent this, we must
```
khuey, line 1344: I think "cause a debug fault before executing the next instruction" is clearer. "exiting an instruction" is not really a well-defined concept, at least in my head.

Keno: Fair point.

src/Task.cc
Keno: Thanks.

Keno:
Also I seem to have broken something in my latest rebase, so I'll fix the two nits and also whatever I broke. One sec.

Keno:
All fixed up.

PR #2599, opened 2020-06-02
Port more tests to aarch64

Keno:


src/replay_syscall.cc
```diff
@@ -1149,14 +1149,14 @@ static void rep_process_syscall_arch(ReplayTask* t, ReplayTraceStep* step,
...
-      r.set_syscallno(t->regs().original_syscallno());
+      r.set_syscallno(sys);
       r.set_ip(r.ip().decrement_by_syscall_insn_length(r.arch()));
       t->set_regs(r);
       if (sys == Arch::mprotect) {
         t->vm()->fixup_mprotect_growsdown_parameters(t);
       }
-      t->exit_syscall();
       t->enter_syscall();
+      t->exit_syscall();
```
khuey, line 1159: This needs some explanation.

Keno: The code was just wrong. exit_syscall advances from an entry stop to an exit stop. enter_syscall advances from any other stop to the entry (or seccomp stop). At this point in the code we just finished a syscall, so we're at an exit stop and want to re-enter it. On x86 this makes little difference, but on aarch64 we need to apply the syscall entry register changes for things to work properly.

Issue #2610, opened 2020-06-22
Intel CPU type 0x706e0 unknown


mdavidsaver:
The CPU is an Intel [i5-1035G1](https://ark.intel.com/content/www/us/en/ark/products/196603/intel-core-i5-1035g1-processor-6m-cache-up-to-3-60-ghz.html).

I have RR 5.2.0 at present (from Debian stable).  Apologies if this has already been added.  It isn't obvious to me which entry would apply.  I would expect to see "Ice Lake" in the list.

https://github.com/mozilla/rr/blob/e89f84c77a31ab6474fdd021517b33ff7cb0a256/src/PerfCounters.cc#L123-L143

khuey:
No, we haven't added Ice Lake yet. It should be fairly straightforward (e.g. https://github.com/mozilla/rr/commit/262e285313fe47c995a73497391f981011975845#diff-fc2123fdd766dab1791fdbcbd769807f, though some of this code moved into src/PerfCounters_x86.h since this changeset). The actual values that go in the `pmu_configs` table are unlikely to have changed since Comet Lake. Can you test and send a PR?

mdavidsaver:
#2611 takes the referenced commit as a template.  However, this gives:

```sh
$ ./build/usr/bin/rr record echo true
[FATAL /home/mdavidsaver/source/rr/src/PerfCounters.cc:277:check_working_counters() errno: ENOENT] 
Got 11 branch events, expected at least 500.

The hardware performance counter seems to not be working. Check
that hardware performance counters are working by running
  perf stat -e r5101c4 true
and checking that it reports a nonzero number of events.
If performance counters seem to be working with 'perf', file an
rr issue, otherwise check your hardware/OS/VM configuration. Also
check that other software is not using performance counters on
this CPU.
=== Start rr backtrace:
./build/usr/bin/rr(_ZN2rr13dump_rr_stackEv+0x35)[0x55fb6b497e1b]
./build/usr/bin/rr(_ZN2rr15notifying_abortEv+0x53)[0x55fb6b497de1]
./build/usr/bin/rr(_ZN2rr12FatalOstreamD1Ev+0x30)[0x55fb6b32f8f8]
./build/usr/bin/rr(+0x380aa0)[0x55fb6b35aaa0]
./build/usr/bin/rr(+0x380bae)[0x55fb6b35abae]
./build/usr/bin/rr(+0x381068)[0x55fb6b35b068]
[...]

PR #2611, opened 2020-06-23
add Intel Icelake

mdavidsaver:
Attempt to address #2610

### DIFF
--- a/src/PerfCounters.cc
+++ b/src/PerfCounters.cc
@@ -66,6 +66,7 @@ enum CpuMicroarch {
   IntelGoldmont,
   IntelKabylake,
   IntelCometlake,
+  IntelIcelake,
   AMDF15R30,
   AMDRyzen,
   ARMNeoverseN1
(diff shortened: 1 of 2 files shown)


src/PerfCounters.cc
```diff
@@ -121,6 +122,7 @@ struct PmuConfig {
 // See Intel 64 and IA32 Architectures Performance Monitoring Events.
 // See check_events from libpfm4.
 static const PmuConfig pmu_configs[] = {
+  { IntelIcelake, "Intel Icelake", 0x5111c4, 0, 0x5301cb, 0, 100, PMU_TICKS_RCB },
```
khuey, line 125: Can we replace 0x5301cb with 0 here? I don't think that counter actually exists on Ice Lake.

mdavidsaver: Done.  Zero seems to make no difference.

khuey: Yeah that entry in the table isn't actually used now.

khuey:
Thanks for the PR and testing!

mdavidsaver:
Thanks for taking the time with this.

PR #2634, opened 2020-08-15
Leave AddressSpace::verify if Task already exited.

bernhardu:
Thanks for the fast response.
Your proposed change fixed the test at my test system.
Fixes 2633.


khuey:
lgtm

PR #2671, opened 2020-09-18
Support for AMD Zen-based processors

### LINKED ISSUE #2034: Support AMD Ryzen?
Ryzen has a conditional branch counter. I have patches to use it here: https://github.com/mozilla/rr/tree/ryzen

To make it work reliably I had to increase the skid counter to 1000. That's pretty high, but OK. The patches make the skid size configurable per-architecture so we don't take that hit on Intel.

With these patches, most tests pass and the rest seem to be intermittent. In one run I get 10 failures out of 2068:
````
        327 - proc_fds (Failed)
        388 - ptracer_death_multithread_peer-no-syscallbuf (Failed)
        688 - condvar_stress-no-syscallbuf (Failed)
        835 - thread_stress (Failed)
        904 - checkpoint_prctl_name-no-syscallbuf (Failed)
        1362 - proc_fds-32-no-syscallbuf (Failed)
        1679 - async_signal_syscalls2-32 (Failed)
        1722 - condvar_stress-32-no-syscallbuf (Failed)
        1869 - thread_stress-32 (Failed)
        1870 - thread_stress-32-no-syscallbuf (Failed)
````

It appears that all these failures are due to intermittent overcounting. In most of them, during recording we seem to have overcounted a few conditional branches in the leadup to some syscall. In the rest, we seem to have overcounted during replay.

One interesting thing is that most of the syscalls where we detect the overcount are an `mprotect` (or a syscall following a syscall-buffered `mprotect`) that followed an `mmap`. There are two exceptions, one a `read` syscall and one a `write` syscall. I need to think about what this might mean.

### PR DESCRIPTION
**Pull request by glandium**
Fixes #2034

### DIFF
--- a/src/PerfCounters_x86.h
+++ b/src/PerfCounters_x86.h
@@ -74,16 +74,16 @@ static CpuMicroarch compute_cpu_microarch() {
 	return IntelCometlake;
     case 0x30f00:
       return AMDF15R30;
-    case 0x00f10:
+    case 0x00f10: // Naples, Whitehaven, Summit Ridge, Snowy Owl (Zen) (UNTESTED)
+    case 0x10f10: // Raven Ridge, Great Horned Owl (Zen) (UNTESTED)
+    case 0x10f80: // Banded Kestrel (Zen), Picasso (Zen+) (UNTESTED)
+    case 0x20f00: // Dali (Zen) (UNTESTED)
+    case 0x00f80: // Colfax, Pinnacle Ridge (Zen+) (UNTESTED)
+    case 0x30f10: // Matisse (Zen 2)
+    case 0x60f10: // Renoir (Zen 2) (UNTESTED)
+    case 0x70f10: // Rome, Castle Peak (Zen 2) (UNTESTED)
       if (ext_family == 8) {
-        if (!Flags::get().suppress_environment_warnings) {
-          fprintf(stderr, "You have a Ryzen CPU. The Ryzen "
-                          "retired-conditional-branches hardware\n"
-                          "performance counter is not accurate enough; rr will "
-                          "be unreliable.\n"
-                          "See https://github.com/mozilla/rr/issues/2034.\n");
-        }
-        return AMDRyzen;
+        return AMDZen;
       }
       break;
     default:
@@ -91,10 +91,7 @@ static CpuMicroarch compute_cpu_microarch() {
   }
 
   if (!strcmp(vendor, "AuthenticAMD")) {
-    CLEAN_FATAL()
-        << "AMD CPUs not supported.\n"
-        << "For Ryzen, see https://github.com/mozilla/rr/issues/2034.\n"
-        << "For post-Ryzen CPUs, please file a Github issue.";
+    CLEAN_FATAL() << "AMD CPU type " << HEX(cpu_type) << " unknown";
   } else {
     CLEAN_FATAL() << "Intel CPU type " << HEX(cpu_type) << " unknown";
   }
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/PerfCounters_x86.h [hunk 1]
```diff
@@ -74,16 +75,16 @@ static CpuMicroarch compute_cpu_microarch() {
 	return IntelCometlake;
     case 0x30f00:
       return AMDF15R30;
-    case 0x00f10:
+    case 0x00f10: // Naples, Whitehaven, Summit Ridge, Snowy Owl (Zen)
```
rocallahan, line 78: Until now we've basically only added lines for archs we have actually tested. I think it's OK to add lines for CPUs we expect to work, but I still want to record which ones have (not) been tested. So please add `(UNTESTED)` for every CPU other than yours that you're adding here.

src/PerfCounters_x86.h
```diff
@@ -265,11 +265,46 @@ static void check_for_xen_pmi_bug() {
...
+  // See more details at https://github.com/mozilla/rr/issues/2034.
+  struct perf_event_attr attr;
+  // 0x25 == RETIRED_LOCK_INSTRUCTIONS - Counts the number of retired locked instructions
+  // + 0x08 == SPECLOCKMAPCOMMIT
+  init_perf_event_attr(&attr, PERF_TYPE_RAW, 0x510825);
+
+  ScopedFd fd = start_counter(0, -1, &attr);
+  if (fd.is_open()) {
+    int atomic = 0;
+    ioctl(fd, PERF_EVENT_IOC_DISABLE, 0);
```
rocallahan, line 283: Why disable the event just before you enable it?

src/PerfCounters_x86.h
```diff
@@ -265,11 +265,46 @@ static void check_for_xen_pmi_bug() {
...
+    ioctl(fd, PERF_EVENT_IOC_ENABLE, 0);
+    int64_t count = read_counter(fd);
+    // A lock add is known to increase the perf counter we're looking at.
+    asm volatile("lock addl $1, %0": "+m" (atomic));
+    if (read_counter(fd) == count) {
+      has_speclockmap_disabled = true;
+    } else {
+      fprintf(stderr,
+              "On Zen CPUs, rr may not work reliably unless you disable the "
```
rocallahan, line 292: Change this to 'will not work reliably...'

rocallahan (changes requested):

src/PerfCounters_x86.h
```diff
@@ -93,7 +94,6 @@ static CpuMicroarch compute_cpu_microarch() {
   if (!strcmp(vendor, "AuthenticAMD")) {
     CLEAN_FATAL()
         << "AMD CPUs not supported.\n"
```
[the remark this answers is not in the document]
rocallahan: er, "*are* not supported"

src/PerfCounters_x86.h
rocallahan: After this has landed we will put out a call for volunteers to test the other microarchs.

src/PerfCounters_x86.h
glandium: I copied this from the other bug checks.

src/PerfCounters_x86.h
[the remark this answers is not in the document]
glandium: > AMD pre-Zen CPUs not supported

AMDF15R30 is supported, though...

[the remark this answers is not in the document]
rocallahan: OK, then let's replace this block with `CLEAN_FATAL() << "AMD CPU type " << HEX(cpu_type) << " unknown";`

src/PerfCounters_x86.h
eddyb: FWIW, my implementation never touches enable/disable (even during regular usage), and doesn't even account for any kind of bias in the count read by `rdpmc`, just relies on *always* subtracting two read counts (or comparing them, like here).

But I'm using `rdpmc` with `perf_event_mmap_page`, not reading from the `fd`, so maybe there's some differences there?

rocallahan (changes requested):

src/PerfCounters_x86.h
rocallahan: Ah, we're just doing the disable thing to reset the counter in case code in `start_counter()` accidentally triggers events. So I think this is fine.

src/PerfCounters_x86.h
```diff
@@ -4,6 +4,7 @@
 static bool has_kvm_in_txcp_bug;
 static bool has_xen_pmi_bug;
 static bool supports_txcp;
+static bool has_speclockmap_disabled;
```
glandium, line 7: actually, this is not necessary.

src/PerfCounters_x86.h
```diff
@@ -265,11 +262,45 @@ static void check_for_xen_pmi_bug() {
...
+  // See more details at https://github.com/mozilla/rr/issues/2034.
+  struct perf_event_attr attr;
+  // 0x25 == RETIRED_LOCK_INSTRUCTIONS - Counts the number of retired locked instructions
+  // + 0x08 == SPECLOCKMAPCOMMIT
+  init_perf_event_attr(&attr, PERF_TYPE_RAW, 0x510825);
+
+  ScopedFd fd = start_counter(0, -1, &attr);
+  if (fd.is_open()) {
+    int atomic = 0;
+    ioctl(fd, PERF_EVENT_IOC_ENABLE, 0);
```
rocallahan, line 280: Sorry --- to be clear, I think we should keep the `PERF_EVENT_IOC_DISABLE`.

rocallahan (changes requested):

src/PerfCounters_x86.h
glandium: It doesn't matter, though, because we read the counter before the lock add, and after, and compare the values. In fact, the check works without PERF_EVENT_IOC_ENABLE.

rocallahan: True. In that case, remove the PERF_EVENT_IOC_ENABLE.

PR #2690, opened 2020-09-20
zen_workaround: Don't enable SSB mitigation when the kernel doesn't allow it

glandium:
It has been confirmed that the workaround sticks when SSB mitigation is
either fully enabled or disabled, as opposed to configurable per
process.

### DIFF
--- a/scripts/zen_workaround.py
+++ b/scripts/zen_workaround.py
@@ -46,25 +46,40 @@ if not args.check:
         os.write(msr, struct.pack('<q', val))
         os.close(msr)
 
+ssb_status = 'unknown'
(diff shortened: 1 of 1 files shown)


glandium:
Note: it was confirmed in the Support AMD Ryzen issue (not linking on purpose to avoid the noise), and I additionally was able to confirm locally by [tweaking my kernel](https://lkml.org/lkml/2020/9/19/328) to allow me to disable the native SSB mitigation and making it behave the same as on Zen/Zen+.

glandium:
@eddyb you may be interested in the patch mentioned above.

PR #2724, opened 2020-10-29
Fix formatting of "rr help record"

SCFrench:
Output of "rr help record" used to contain this snippet:

```
  --asan                     Override heuristics and always enable ASAN
                             compatibility.Global options:
  --disable-cpuid-faulting   disable use of CPUID faulting
```

Now it looks like this:

```
  --asan                     Override heuristics and always enable ASAN
                             compatibility.
Global options:
  --disable-cpuid-faulting   disable use of CPUID faulting
```


khuey:
Thanks!

PR #2725, opened 2020-10-29
Don't run tests marked with skip_if_32_bit at i386.

bernhardu:


bernhardu:
Please wait, I have some test failures that seem to be caused by this ...

bernhardu:
The first version was too much as bitness is used in other places too.
This iteration moved the check into the function skip_if_32_bit.

PR #2754, opened 2020-12-01
rm leftover rr_page install rules

Keno:
These were superseded by librrpage.so.

@yuyichao see if this fixes your build issue.


yuyichao:
Yes it seems to work. Thx.

PR #2755, opened 2020-12-01
Make sure to use the proper /proc/<pid>/exe when detaching

Keno:
This is useful in case the child expectes to be able to exec
`/proc/self/exe`. I originally thought it would be possible to
set this using PR_SET_MM, like we do the rest of the MM settings,
but turns out that this operation is behind a capabilities check
(in fact, there are two code paths to this function that have
different capabilities checks). Just use the correct exe up
front, so we don't have this problem.

This is the first change that uses the rrcalls directly from
a test file. Arguably that means they should move to rr.h, but
I don't quite feel that the detaching mechanism is robust enough
yet to consider it part of the "public" API (and maybe the "public"
API shouldn't be a raw syscall, but rather something exported from
the preload library).


yuyichao:
This seems to break compilation for the 32bit binary?

```
In file included from /build/rr-git/src/rr/build/32/detach_state.c:3:
/build/rr-git/src/rr/build/32/util_internal.h:7:10: fatal error: rrcalls.h: No such file or directory
    7 | #include "rrcalls.h"
      |          ^~~~~~~~~~~
compilation terminated.
```

Keno:
Ah, sorry I had switched from also copying that header to using `target_include_directories`, but looks like I still had the header copied locally in the build directory, so it worked anyway. Fix coming up.

khuey:
I hit it as well. Fix is pushed.

Keno:
Thanks. Sorry about that. I guess we should try to setup CI again.

khuey:
No worries. It's vaguely on our todo list now that we have control of the repo but it's not a high priority.

PR #2756, opened 2020-12-01
Explicitly set offsets in rr_page linker script

Keno:
Apparently the binutils on Ubuntu 18.04 has a very large value of
MAXPAGESIZE. Just hardcode a 4k page instead. This might break on
aarch64, but we're not using the syscallbuf there yet anyway, so
let's revisit that when we get there.

@khuey See if this fixes things for you. Seems to work locally with 18.04.


khuey:
Yes, that does it, thank you.

PR #2761, opened 2020-12-04
32bit fixes related to 3aaf7920e11

bernhardu:
Hello, tried to run the tests on plain i386 and found most tests did fail.
With the libname in map_rr_page fixed just two tests were then failing,
therefore I wondered if the tests do not run at amd64 as _32 and tried
to change this.
Please note that the change to is_rr_page_lib got not triggered
by any issue, just finding me there just one of the libname.

The test record_replay still fails by hanging infinitely at i386,
but not sure if this is related to rrpage or not.

cc: @Keno


Keno (approved):

bernhardu:
Hello @khuey, sorry for my impatience, are there still issues with these patches?

khuey:
Nope, we're good to go, thanks for the fixes.

PR #2783, opened 2021-01-04
WIP: Add support to run tests with 32-bit rr below 64-bit kernel (prctl_mm_map)

bernhardu:
This helps tests detach_state, nested_detach and nested_detach_wait
to succeed with a 32-bit rr running under a 64-bit kernel.

This iteration of the series has moved the structure definition to kernel_abi.h and replaced all uses of sys/prctl.h's definition like @khuey asked for.

Also I hope this should make my use of fflush in the commit message in the last iteration more clear.
Is it known what this fflush changes to make the prctl not to fail with EPERM?

### DIFF
--- a/src/Task.cc
+++ b/src/Task.cc
@@ -14,7 +14,6 @@
 #include <stdlib.h>
 #include <string.h>
 #include <sys/personality.h>
-#include <sys/prctl.h>
 #include <sys/resource.h>
 #include <sys/socket.h>
 #include <sys/time.h>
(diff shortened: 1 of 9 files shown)


src/Task.cc
```diff
@@ -3448,12 +3447,55 @@ static void create_mapping(Task *t, AutoRemoteSyscalls &remote, const KernelMapp
...
-                            sizeof(map));
+#if defined(__x86_64__)
+  /* This fflush somehow makes the following prctl succeed
+     with a 64-bit rr under 64-bit kernel instead of giving a EPERM error.
+     What does that change to our process? Why is this not needed for x86? */
+  fflush(stderr);
+#endif
+
+  unsigned int expected_size = 0;
+  int result = prctl(PR_SET_MM, PR_SET_MM_MAP_SIZE, &expected_size);
```
Keno, line 3460: Unused arguments should be specified as zero, otherwise this prctl can fail. (may explain what the previous fflush accomplishes)

bernhardu: That's it - giving zwo more zero arguments make the prtcl succeed. Thank you very much.

Keno (approved):

PR #2791, opened 2021-01-14
Fix for ancient libc build

Keno:


khuey:
Yikes that's old.

Keno:
Our (julia) platform support policy includes CentOS 6, which has libc from 2010, so we test this path in our deployment scripts.

khuey:
How does that have a kernel new enough for rr?

Keno:
Usually people run it in a docker container on a modern kernel.

khuey:
Ah.

PR #2813, opened 2021-02-16
Shove any .note.gnu.property sections up into the header along with the other notes

khuey:
@sefeng reported an issue where this was dislocating .vdso.text with a .note.gnu.property section with gcc 10.2 and ld 2.36.

This fixes his immediate problem, but this is still not robust to other .note sections (or perhaps even other types of sections). Any thoughts @rocallahan @Keno

### DIFF
--- a/src/preload/rr_page.ld
+++ b/src/preload/rr_page.ld
@@ -13,6 +13,7 @@ SECTIONS
   .eh_frame_hdr   : { *(.eh_frame_hdr) *(.eh_frame_entry .eh_frame_entry.*) } :header :eh_frame
   .eh_frame       : { KEEP (*(.eh_frame)) *(.eh_frame.*) } :header :eh_frame
   .note.gnu.build-id  : { *(.note.gnu.build-id) } :header :note
+  .note.gnu.property : { *(.note.gnu.property) } :header :note
   .hash           : { *(.hash) } :header
   .gnu.hash       : { *(.gnu.hash) } :header
   .dynsym         : { *(.dynsym) } :header


rocallahan (approved):
Looks good to me

sefeng211:
Not sure who was tagged hehe, but thanks for the patch!

Keno:
LGTM. I'm not sure there's a way to specify that all note section are ok to include and where to put them. We can certainly ask the linker to discard them, but that's probably not what we want.

khuey:
Is there a way we can make the build fails if it produces a librrpage.so that's going to fail when we do our not-quite-dl-compatible load of it on exec?

Keno:
I assume what happened here is that it decided to just dump that section at the end of the file, which broke things? I'm thinking there should be a way to have it error on unknown sections. Let me see.

Keno:
If you have a test case, maybe see if you can make something like this work?
```
diff --git a/src/preload/rr_page.ld b/src/preload/rr_page.ld
index 568bde29..caf60ce6 100644
--- a/src/preload/rr_page.ld
+++ b/src/preload/rr_page.ld
@@ -35,4 +35,9 @@ SECTIONS
   . = 0x70001000;
   .replay.text : { *(.replay.text) } :replay
   /DISCARD/ : { *(.debug_* ) }
+  .rest : {
+    PROVIDE (__rest_start = .);
+    *(*) 
+    ASSERT(__rest_start == ., "Error: Unhandled section in librrpage linker script");
+  }
 }
```

khuey:
@sefeng211 can you try that (without the .note.gnu.property patch) and see if it errors out?

sefeng211:
I applied the above patch to fd11f49f943f13bdd0a158735f610ad4cd183348, and wasn't able to compile. 
```
[  3%] Linking CXX executable bin/test-monitor
[  3%] Linking C shared library lib/rr/librrpreload.so
[  3%] Built target test-monitor
[  3%] Built target rrpreload
[  3%] Linking C shared library lib/rr/librrpreload_32.so
[  3%] Built target rrpreload_32
[  3%] Linking C static library libbrotli.a
[  3%] Built target brotli
make: *** [Makefile:182: all] Error 2
```

I am happy to do more testing.

PR #2830, opened 2021-03-30
Fix missing KEYCTL_GET_PERSISTENT operation for keyctl syscall.

vonasek:
This fixes crash when running:
$ rr keyctl get_persistent @p


khuey:
thanks!

PR #2926, opened 2021-07-14
Fixes build on Ubuntu 16.04

bernhardu:
Tried to run the tests in a rr enabled VM with Ubuntu 16.04, but these triggered
several times a kernel issue `kernel BUG at /build/linux-3Zqs18/linux-4.4.0/mm/memory.c:3214!`.
(With `4.4.0-210-generic`, `Description:    Ubuntu 16.04.7 LTS`)
Having now the first 1300 tests run, 95% did succeed.

Doing a build and test run on my usual setup, all tests succeed with these two patches.

Related to issue: #2919

### DIFF
--- a/src/kernel_supplement.h
+++ b/src/kernel_supplement.h
@@ -421,6 +421,10 @@ enum {
 #define CAP_PERFMON 38
 #endif
 
+#ifndef SEGV_PKUERR
+#define SEGV_PKUERR 4
+#endif
+
 } // namespace rr
 
 #endif /* RR_KERNEL_SUPPLEMENT_H_ */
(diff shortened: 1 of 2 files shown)


khuey:
If rr crashes this kernel should we really be fixing the build on it?

bernhardu:
(It finished now, the remaining part of the test run did better, the end result is: `97% tests passed, 73 tests failed out of 2673`.)


> If rr crashes this kernel should we really be fixing the build on it?

The VM stayed usable, but there are now a bunch of processes in state D and Z.
And maybe it could be of value for (the few) people either running newer kernels with this ancient user land,
or their payload happens not to make rr trigger these issues.

PR #2978, opened 2021-10-10
audit/stap-note-iter: map ELF header directly

Bob131:
In order to dig around in a loaded object's STap probe sections,
stap_note_iter_init() tries to find the address at which the ELF
header was mapped by the linker/kernel; to do this, it uses a function
private to glibc (_dl_addr) to find the base address of an object in
memory given the provided pointer to its dynamic segment.

At some point (apparently due a toolchain update between F34 and F35,
but I'm not entirely sure) this stopped working because _dl_addr() is
no longer being exported in libc's DT_GNU_HASH table: the execution of
stap_note_iter_init() triggers an exit when the PLT stub fails to
resolve the symbol.

Rather than using some other functionality or poking around trying to
find where the header might already be mapped, this commit has the
function just map it as required. (I have no idea why I wrote it like
this in the first place; I suspect I was being too clever by half.)

### DIFF
--- a/src/audit/stap-note-iter.c
+++ b/src/audit/stap-note-iter.c
@@ -14,17 +14,6 @@
 #define ALIGN_UP(x, p2) \
   (((x) & ((p2) - 1)) == 0 ? (x) : ((x) + (p2)) & ~((p2) - 1))
 
-/* rtld doesn't mark itself as initialised until after it's loaded all the
- * initially required objects. dladdr() checks this flag, and when not set it
- * delegates functionality to dlfcn_hook; in our case no such hook is
- * installed, so it segfaults trying.
- *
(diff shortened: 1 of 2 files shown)


Bob131:
Running the test suite revealed a failure of `jit_proc_mem` when `--stap-sdt` is enabled, which appears to be a manifestation of #2163. It seems as though the audit library triggers a call to `rr::ReplayTimeline::singlestep_with_breakpoints_disabled()` when reverse-continuing in a way that the normal rtld startup doesn't, and rr dies trying to remove a breakpoint in memory that's yet to be mapped.

I don't think this can be meaningfully addressed by this patch (since it's not really specific to the SysTap machinery), but I thought I'd just mention it here for future reference.

PR #2989, opened 2021-11-12
allow compilation without VIDIOC_QUERY_EXT_CTR

GitMensch:
fixing build issues with old environments after af1e9b5

### DIFF
--- a/src/record_syscall.cc
+++ b/src/record_syscall.cc
@@ -1942,7 +1942,9 @@ static Switchable prepare_ioctl(RecordTask* t,
     case IOCTL_MASK_SIZE(VIDIOC_G_OUTPUT):
     case IOCTL_MASK_SIZE(VIDIOC_S_CTRL):
     case IOCTL_MASK_SIZE(VIDIOC_G_INPUT):
+    #ifdef VIDIOC_QUERY_EXT_CTRL
     case IOCTL_MASK_SIZE(VIDIOC_QUERY_EXT_CTRL):
+    #endif
     case IOCTL_MASK_SIZE(VIDIOC_G_PRIORITY):
     case IOCTL_MASK_SIZE(VFAT_IOCTL_READDIR_BOTH):
       syscall_state.reg_parameter(3, size, IN_OUT);
--- a/src/test/video_capture.c
+++ b/src/test/video_capture.c
@@ -59,6 +59,7 @@ static int open_device(void) {
     atomic_printf("%s VIDIOC_G_INPUT returns %d\n", device_name, input);
   }
 
+  #ifdef VIDIOC_QUERY_EXT_CTRL
   struct v4l2_query_ext_ctrl qec;
   memset(&qec, 0, sizeof(qec));
   qec.id = V4L2_CTRL_FLAG_NEXT_CTRL | V4L2_CTRL_FLAG_NEXT_COMPOUND;
@@ -69,6 +70,7 @@ static int open_device(void) {
     atomic_printf("%s VIDIOC_QUERY_EXT_CTRL returns id=%d, type=%d, name=%s\n",
                   device_name, qec.id, qec.type, qec.name);
   }
+  #endif
 
   enum v4l2_priority prio = V4L2_PRIORITY_UNSET;
   ret = ioctl(fd, VIDIOC_G_PRIORITY, &prio);


khuey:
Is there a reason not to define this ioctl constant in kernel_supplement.h (i.e. what we do for `HCIGETDEVLIST`)?

GitMensch:
Thanks for the quick answer!

I don't know of one (just did not know about the supplement file), but assume that because of the special video tests this _may_ not work.

If that supplement is fine then it is the nicer approach - feel free to close this issue and adjust the supplement file in this case.

khuey:
Is `struct v4l2_query_ext_ctrl` defined on your system?

GitMensch:
Doesn't look like that:

~~~
/tmp/rr-5.5.0/src/test/video_capture.c: In function 'open_device':
/tmp/rr-5.5.0/src/test/video_capture.c:62:30: error: storage size of 'qec' isn't known
   62 |   struct v4l2_query_ext_ctrl qec;
      |

bernhardu:
> but assume that because of the special video tests this _may_ not work.

Could both be done? Defining VIDIOC_QUERY_EXT_CTRL in kernel_supplement.h to have it inside rr binary, and ignore this part of the test if VIDIOC_QUERY_EXT_CTRL is not defined in the system?

khuey:
Yeah that's where I was going with this but rr would also need to bundle a definition of `struct v4l2_query_ext_ctrl` apparently.

bernhardu:
Now I see, the macro definition uses the struct. Had not looked into it deep enough before, sorry for the noise.

khuey:
I think ideally we'd define VIDIOC_QUERY_EXT_CTRL and the struct layout in kernel_supplement.h if VIDIOC_QUERY_EXT_CTRL is not already defined. The test can be left with #ifdef VIDIOC_QUERY_EXT_CTRL.

@GitMensch do you want to implement that?

GitMensch:
Why should we leave the `#ifdef` in when we define it?

khuey:
To be clear, I mean leave the #ifdef in the tests. The tests don't include kernel_supplement.h and if the headers don't include this ioctl it's very unlikely the kernel on the machine will support it.

GitMensch:
OK, then we don't need to define the struct either...

khuey:
Yeah, we do, because

```
khuey@minbar:~/dev/rr$ grep VIDIOC_QUERY_EXT_CTRL -r /usr/include/
/usr/include/linux/videodev2.h:/*  Used in the VIDIOC_QUERY_EXT_CTRL ioctl for querying extended controls */
/usr/include/linux/videodev2.h:#define VIDIOC_QUERY_EXT_CTRL    _IOWR('V', 103, struct v4l2_query_ext_ctrl)
```

Take a look at what we do for EVIOCGMASK.

khuey:
Perhaps that doesn't really matter though since we ultimately use `IOCTL_MASK_SIZE(EVIOCGMASK)`. It should be pretty easy to just define our own version of the struct and get the right size though.

GitMensch:
I'm too confused what to do now :-(

@khuey Do you mind taking over fixing #2993?

bernhardu:
Hello @GitMensch , if I understood it right the part from your patch about src/test/video_capture.c is fine.
But in src/record_syscall.cc the #ifdef should be dropped.
Therefore one of the following should be one of the following.

```diff
--- a/src/kernel_supplement.h
+++ b/src/kernel_supplement.h
@@ -425,6 +425,13 @@ enum {
 #define SEGV_PKUERR 4
 #endif
 
+#ifndef VIDIOC_QUERY_EXT_CTRL
+#define VIDIOC_QUERY_EXT_CTRL _IOC(_IOC_READ|_IOC_WRITE,'V',103,0)
+/* This definition omits the size because in prepare_ioctl
+   it is masked away anyway. And the real size is taken from
+   the real request by _IOC_SIZE(request). */
+#endif
+
 } // namespace rr
 
 #endif /* RR_KERNEL_SUPPLEMENT_H_ */
```



```diff
--- a/src/kernel_supplement.h
+++ b/src/kernel_supplement.h
@@ -425,6 +425,11 @@ enum {
 #define SEGV_PKUERR 4
 #endif
 
+#ifndef VIDIOC_QUERY_EXT_CTRL
+struct rr_v4l2_query_ext_ctrl { char dummy[232] };
+#define VIDIOC_QUERY_EXT_CTRL _IOWR('V', 103, struct rr_v4l2_query_ext_ctrl)
+#endif
+
 } // namespace rr
 
 #endif /* RR_KERNEL_SUPPLEMENT_H_ */
```



```diff
--- a/src/kernel_supplement.h
+++ b/src/kernel_supplement.h
@@ -425,6 +425,27 @@ enum {
 #define SEGV_PKUERR 4
 #endif
 
+#ifndef V4L2_CTRL_MAX_DIMS
+#define V4L2_CTRL_MAX_DIMS (4)
+#endif
+#ifndef VIDIOC_QUERY_EXT_CTRL
+struct v4l2_query_ext_ctrl {
+  uint32_t id;
+  uint32_t type;
+  char name[32];
+  int64_t minimum;
+  int64_t maximum;
+  uint64_t step;
+  int64_t default_value;
+  uint32_t flags;
+  uint32_t elem_size;
+  uint32_t elems;
+  uint32_t dims[V4L2_CTRL_MAX_DIMS];
+  uint32_t reserved[32];
+};
+#define VIDIOC_QUERY_EXT_CTRL _IOWR('V', 103, struct v4l2_query_ext_ctrl)
+#endif
+
 } // namespace rr
 
 #endif /* RR_KERNEL_SUPPLEMENT_H_ */
```

The first variant has the downside that the define is not really the same value, but should be sufficient for now.
But I am not sure which variant is the preffered one.

khuey:
Please test what I pushed and let me know if there are still problems.

GitMensch:
Thats much more clean than our changes. Thanks, I'll try to give it a test over the week.

khuey:
I just committed the best of what the two of you suggested :)

bernhardu:
Just for the record, did not knew that sizeof(0) is 4 - but doesn't matter. Thanks for fixing.

```
#define IOCTL_MASK_SIZE(v) ((v) & ~(_IOC_SIZEMASK << _IOC_SIZESHIFT))
#define xVIDIOC_QUERY_EXT_CTRL _IOC(_IOC_READ|_IOC_WRITE,'V',103,0)
#define yVIDIOC_QUERY_EXT_CTRL _IOWR('V', 103, 0)
atomic_printf("x=%x y=%lx masked=%lx real=%lx\n", xVIDIOC_QUERY_EXT_CTRL, (long unsigned int)yVIDIOC_QUERY_EXT_CTRL, (long unsigned int)IOCTL_MASK_SIZE(VIDIOC_QUERY_EXT_CTRL), VIDIOC_QUERY_EXT_CTRL);
```

```
x=c0005667 y=c0045667 masked=c0005667 real=c0e85667
```

GitMensch:
Tested, works like a charm.

PR #3029, opened 2022-01-06
Fix build on old glibc

Keno:
Courtesy of the julia deployment system which wants to keep compatibility
with CentOS6 for certain HPC users ;). This is for the `SEEK_HOLE`/`SEEK_DATA`
definitions. Builds fine locally on modern libc also.

### DIFF
--- a/src/EmuFs.cc
+++ b/src/EmuFs.cc
@@ -4,6 +4,7 @@
 
 #include <syscall.h>
 #include <sys/mman.h>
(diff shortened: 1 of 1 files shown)


GitMensch:
Wow, CentOS6 - that's interesting! What kernel is used there - and are there more adjustments to fix compatibility issues?

Keno:
Generally these are modern kernels with old userspace for compatibility.

PR #3042, opened 2022-01-19
Handle backwards compatibility for traces without force_tick.

### LINKED ISSUE #3041: Assumption that no ticks passing between a syscallbuf flush and the succeeding event means the process was SIGKILLed is not backwards compatible
We assume that if no ticks pass between a syscallbuf flush and the succeeding event, the succeeding event represents a SIGKILL and is safe to skip. https://github.com/rr-debugger/rr/blob/f99198bc5eb223065de398beced668d2bc801036/src/ReplaySession.cc#L1393-L1399

This is true with the current version of rr because of `force_tick`, but that is not present in older `librrpreload.so`s. This change is not backwards compatible. A Pernosco customer hit this when recording with an rr from June and then uploading to Pernosco (which is on tip). The trace appears to call ptrace(PTRACE_PEEKDATA) repeatedly without intervening conditional branches. Those are handled by the syscallbuf but at some point it fills up and triggers a flush, and then no ticks pass before the next syscall, leading the current rr to believe it can skip the next event. Divergence ensues.

khuey:
Fixes #3041

### DIFF
--- a/src/rr_trace.capnp
+++ b/src/rr_trace.capnp
@@ -110,6 +110,10 @@ struct Header {
     # Whether the version of rr that recorded this saved the extra registers
     # for the pkey_alloc syscall.
     pkeyAllocRecordedExtraRegs @20 :Bool = false;
+
+    # Whether the version of rr that recorded this forced a tick after each
+    # buffered syscall.
+    bufferedSyscallForcedTick @21 :Bool = false;
   }
   # Are we known to be in chaos mode? Useful for debugging.
   chaosMode @16 :ChaosMode = unknown;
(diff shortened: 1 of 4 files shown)


src/rr_trace.capnp
```diff
@@ -110,6 +110,10 @@ struct Header {
     # Whether the version of rr that recorded this saved the extra registers
     # for the pkey_alloc syscall.
     pkeyAllocRecordedExtraRegs @20 :Bool = false;
+
+    # Whether the version of rr that recorded this forced a tick after each
+    # buffered syscall.
+    bufferedSyscallForcedTick @21 :Bool = false;
```
Keno, line 116: I like it better if the `quirk` is true in the old version and false in the new version, such that a trace produced by a current version of rr has no "quirks", but that's a minor preference.

Keno (approved):

src/rr_trace.capnp
khuey: I actually agree with you but roc wanted it the other way when I did the pkey_alloc(2) quirk above so that there wasn't a negative in the quirk (e.g. "bufferedSyscallDidntForceTick").

khuey:
Thanks for the review.

PR #3043, opened 2022-01-19
Set up Buildkite CI on the `juliaecosystem` queue provided by the Julia Project

DilumAluthge:
In the Julia Project, we use Buildkite[^1] to run CI on our own machines. Most of our machines are reserved for use by Base Julia CI on the main Julia repository[^2]. However, we do have one Buildkite queue (the `juliaecosystem` queue) that is intended for use by repositories in the Julia ecosystem.

We thought it might be nice to use the `juliaecosystem` queue to offer some of our CI resources to the rr project.

[^1]: https://buildkite.com/julialang
[^2]: https://github.com/JuliaLang/julia

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,0 +1,18 @@
+steps:
+  - label: "Build and test"
+    agents:
+      os: "linux"
+      queue: "juliaecosystem"
+    timeout_in_minutes: 60
+    commands: |
+      echo "Print kernel information"
+      uname -a
+      echo "--- Generate build environment"
+      rm -rf obj
+      mkdir obj
+      cd obj
+      cmake ..
+      echo "--- Build"
+      make -j8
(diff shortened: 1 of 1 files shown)


Keno:
@khuey @rocallahan For context, @DilumAluthge and the Julia infrastructure team have offered to extend our CI to rr. These machines are running rr anyway, so they'll be properly configured to run rr and adding a new repo is pretty easy. I'm assuming we'll want to take advantage of this, but let me know if there are any concerns.

khuey:
Sure, sounds great. What else do we need to do?

Keno:
@DilumAluthge said I need to toggle some settings (needs to be somebody who has admin permissions both on the repo and the CI system). Let me try to do that.

vchuravy:
It should be noted that these machines run on AMD EpyC processors, if needed we can add some older Intel CPUs into the mix as well.

PR #3044, opened 2022-01-20
Fix the CI configuration

DilumAluthge:


Keno:
Probably should run `--output-on-failure` and ideally also turn the leftover test directories into an artifact for debugging. The test script puts those in $TMPDIR, so we should probably just point that to some directory that we zip up afterwards. Successful tests should delete their working directories, so a passing build should not have failures.

Keno:
But regardless, looks like CI worked, though we have some failures. I'll look into those, but I think they're probably real, so I'll merge this as is.

PR #3046, opened 2022-01-20
CI: run tests with `ctest --output-on-failure`

DilumAluthge:


DilumAluthge:
I'll make a separate PR for the "upload debug artifact" step.

DilumAluthge:
@Keno Take a look at the logs and make sure that the desired output is being shown when a test fails.

Keno:
Yes, that's great thanks. Out of curiosity, does buildkite have any facility for reading the test summary file that ctest spits out rather than just the plain pass/fail? I know we're not using that in Julia land, but I figured it might have some standard thing.

Keno:
Another thing that ctest does when running locally is remember how long each test takes and run the longest running test first to maximize parallelism. This helps a lot on the rr test suite, because we have a couple of very long running tests. Unfortunately, I'm not sure where it keeps that information, but it'd be great if we could persist it (or maybe we could check it in occasionally - it doesn't have to be perfect, just give ctest a good test order for parallelism).

DilumAluthge:
What format does ctest output?

I think they have some stuff for JUnit, but I'm not sure about the ctest format. @staticfloat have you run across anything for the ctest format for Buildkite?

Keno:
If ctest is recent enough (past couple months), it has --output-xunit which outputs JUnit format.

DilumAluthge:
The two ways to pursue this would be:

1. Find a Buildkite plugin that directly parses whatever file format is output by ctest.
2. Figure out how to get ctest to output in a different format (JUnit or XUnit or whatever). And then find a Buildkite plugin that parses JUnit/XUnit/whatever.

DilumAluthge:
Ah you beat me to it. Let's try the `--output-xunit` approach.

We have full control over the rootfs image, so we can install a newer CMake if need be.

DilumAluthge:
> (or maybe we could check it in occasionally - it doesn't have to be perfect, just give ctest a good test order for parallelism).

This will probably be the quickest way, so I'd say let's start with that. Persisting the info on the Buildkite agents is not impossible, but will require a bit more work than just checking a file into source control with the suggested test run order.

DilumAluthge:
> If ctest is recent enough (past couple months)

From the log[^1], looks like we installed cmake 3.13.4:

```
I: Retrieving cmake 3.13.4-1
I: Validating cmake 3.13.4-1
I: Retrieving cmake-data 3.13.4-1
I: Validating cmake-data 3.13.4-1
```

Is that recent enough? I think that's what Debian Buster gives you when you do `apt-get install cmake`.

[^1]: https://github.com/JuliaCI/rootfs-images/runs/4866729616?check_suite_focus=true

Keno:
> This will probably be the quickest way, so I'd say let's start with that. Persisting the info on the Buildkite agents is not impossible, but will require a bit more work than just checking a file into source control with the suggested test run order.

Ok, looks like test file in question is `CTestCostData.txt` and needs to be put in `$buildir/Testing/Temporary/CTestCostData.txt`. Should we check it into `.buildkite` and then just move it into place?

Here's the one from my local machine:

[CTestCostData.txt](https://github.com/rr-debugger/rr/files/7902075/CTestCostData.txt)

Keno:
> Is that recent enough?

No, that's 2 years old. Looks like we need 3.21

Keno:
Also looks like we're getting duplicate status posts?

DilumAluthge:
> Also looks like we're getting duplicate status posts?

Yeah I'll fix the duplicates.

DilumAluthge:
> Ok, looks like test file in question is `CTestCostData.txt` and needs to be put in `$buildir/Testing/Temporary/CTestCostData.txt`. Should we check it into `.buildkite` and then just move it into place?

That sounds good to me!

Keno:
@DilumAluthge Looking at the `dev_zero` failures, looks like the rootfs may be missing a `/dev/zero` entry? Might be a general Sandbox.jl issue, since in the alpine rootfs I only see:

```
/ # ls -la /dev
total 0
drwxrwxr-x    1 root     root           160 Jan 20 04:08 .
drwxrwxr-x    1 root     root            60 Jan 20 04:08 ..
crw-rw-rw-    1 nobody   nobody      1,   3 Jan  8  2021 null
crw-rw-rw-    1 root     root        5,   2 Jan 20 04:08 ptmx
drwxr-xr-x    2 root     root             0 Jan 20 04:08 pts
drwxrwxrwt    3 nobody   nobody          60 Jan 20 03:41 shm
crw-rw-rw-    1 nobody   nobody      5,   0 Jan 20 03:33 tty
crw-rw-rw-    1 nobody   nobody      1,   9 Jan  8  2021 urandom
```

Keno:
> Looking at the `dev_zero` failures, looks like the rootfs may be missing a `/dev/zero` entry?

Also explains the `clone` failure and a few of the timeouts.

DilumAluthge:
> Looking at the `dev_zero` failures, looks like the rootfs may be missing a `/dev/zero` entry?

@staticfloat How might we go about remedying this?

Keno:
> @staticfloat How might we go about remedying this?

Assuming it still works the way it did when I wrote it, you need to add it here: https://github.com/JuliaPackaging/Yggdrasil/blob/8be19820afa24ff817185f3c53c4d20786a8add3/0_RootFS/Rootfs/bundled/utils/sandbox.c#L345-L347

DilumAluthge:
Just FYI, for the version we use on Buildkite, we use the one bundled with Sandbox.jl: https://github.com/staticfloat/Sandbox.jl/blob/main/deps/userns_sandbox.c

DilumAluthge:
https://github.com/staticfloat/Sandbox.jl/pull/70

Keno:
@DilumAluthge This PR by itself seems good to go. Shall I merge it or will that conflict with your other PRs?

DilumAluthge:
Go ahead and merge; if there are merge conflicts, they shouldn't be too hard to resolve.

DilumAluthge:
I've resolved the merge conflicts.

PR #3048, opened 2022-01-20
CI: Upload a tarball of the `TMPDIR` as a Buildkite artifact

DilumAluthge:


.buildkite/pipeline.yml
```diff
@@ -14,6 +14,10 @@ steps:
           rootfs_treehash: "8b5c776db01b54e8632c31fece119b5fe7d86f91"
           # workspaces:
           #   - "/cache/repos:/cache/repos"
+      - JuliaCI/julia#v1:
```
Keno, line 17: Why is this in here twice?

DilumAluthge: The order of operations is:

1. Install Julia in the outer sandbox.
2. Set up the inner sandbox (which requires Julia to set up) and enter the inner sandbox.
3. Install Julia in the inner sandbox.

The inner sandbox is the one that has all the tools (e.g. cmake) from our custom rootfs image. The outer sandbox is very minimal and basically just has the Buildkite agent.

Keno: I see. It's Julia all the way down.

DilumAluthge: Until you reach turtles.

DilumAluthge:
Okay, we are now successfully uploading a tarball of `TMPDIR` as a Buildkite artifact. To download the artifact, go to the Buildkite log, click on the job, and then click on the "Artifacts" tab.

@Keno Can you take a look at the tarball from the most recent job, and let me know if it looks like the tarball includes the files that you wanted it to?

Also, right now we are only uploading the `TMPDIR` directory, but if there are other files and/or directories that you want us to upload, it's easy enough to add them.

Keno:
Will check in the morning

Keno:
> @Keno Can you take a look at the tarball from the most recent job, and let me know if it includes the correct files?

LGTM. We could do Julia CI style trace compression to get the size of these down a bit, but of course in theory the size of these should be 0 if no tests fail, so let's just leave it as is.

DilumAluthge:
I'll fix the merge conflicts and then ping you when CI is green.

DilumAluthge:
@Keno I've fixed the merge conflicts, so this is good to go.

PR #3050, opened 2022-01-20
CI: Move the `CTestCostData.txt` file into place before running the tests

DilumAluthge:
This uses the `CTestCostData.txt` file from Keno's comment here: https://github.com/rr-debugger/rr/pull/3046#issuecomment-1017086812


Keno:
I think this is working. It's starting with the long-running tests. We may see less benefit while we still have failing tests, since some of them are timeouts, which makes the cost values inaccurate.

DilumAluthge:
Yeah, once we have fixed some of the failing tests, it will probably be worth updating the cost file.

Keno:
I think this is working, but my local test file worked less well than I expected `vsyscall_reverse_next-no-syscallbuf` takes no time locally, but almost 300s on CI (and got scheduled right at the end). Not sure what the difference is. But regardless, let's merge this, fix all the tests and then generate a cost file on CI that matches that environment.

PR #3051, opened 2022-01-20
CI: remove the duplicate commit status

DilumAluthge:
In the Buildkite settings[^1], I've enabled the "Update commit statuses" option. So we don't need to also specify it in the pipeline YAML.

[^1]: https://buildkite.com/julialang/rr/settings/repository


DilumAluthge:
Looks like this has correctly fixed the duplicate commit statuses, i.e. there is now only a single commit status.

PR #3054, opened 2022-01-20
CI: update to `cmake` version 3.22.1

DilumAluthge:
This will allow us to use the `ctest --output-junit <file>` option in the future.


DilumAluthge:
We should probably revert this 😭 

Look at the log.

Keno:
That's what I get for merging things late at night. I think it's fine, just put up a new PR that fixes the issue. CI will be broken until then, but that's no different from what the situation was earlier today.

PR #3059, opened 2022-01-20
CI: Print CPU information at the beginning of the job

DilumAluthge:


DilumAluthge:
@Keno @vtjnash

DilumAluthge:
Ah, `lscpu` doesn't work inside the sandbox, which is probably a good think.

I don't think it's worth trying to get `lscpu` working inside the sandbox.

Instead, if a job is failing, and someone wants to know which CPU that job ran on, just ping me or @Keno - we can SSH into the machine and run `lscpu` outside the sandbox, and then provide that information to you.

khuey:
Seems like you ought to be able to run `cpuid -1`, or just `cat /proc/cpuinfo` or something.

DilumAluthge:
`cpuid -1` doesn't work inside the sandbox, but `cat /proc/cpuinfo` does!

khuey:
`awk -v RS= 'NR==1' /proc/cpuinfo` is probably better than straight `cat` since it'll avoid repeating for each CPU

DilumAluthge:
@Keno This one is also good to go.

DilumAluthge:
Bump @Keno

PR #3060, opened 2022-01-21
CI: for now, skip the tests that are known to fail; we will incrementally add them back as we fix them

DilumAluthge:


Keno:
Can we fix the `/dev/zero` thing first? That'll fix most of these, so the list will be more much more managable.

DilumAluthge:
> Can we fix the `/dev/zero` thing first?

We'll need https://github.com/JuliaPackaging/Yggdrasil/pull/4294

Keno:
All tests pass on master, so this is unnecessary now.

PR #3061, opened 2022-01-21
CI: pass `verbose: true` to the sandbox plugin

DilumAluthge:
This allows us to see what versions of `Sandbox.jl` and `UserNSSandbox_jll.jl` are being used.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -12,6 +12,7 @@ steps:
       - staticfloat/sandbox#v1:
           rootfs_url: https://github.com/JuliaCI/rootfs-images/releases/download/v4.17/rr.x86_64.tar.gz
           rootfs_treehash: "838680473e6ffd8977e1957904d418640e75f69e"
+          verbose: true
       - JuliaCI/julia#v1:
           persist_depot_dirs: packages,artifacts,compiled
           version: '1.6'


DilumAluthge:
@Keno I just triggered a new build, which is using the new sandbox with the `/dev/zero` fix. When it finishes, let's take a look and see how many failures remain.

DilumAluthge:
It timed out at 30 minutes. I'll increase the timeout, but it's a little concerning that it's taking so long.

Keno:
Yes, I don't think those are ever coming back. For reference, the tests that timed out were:
```
          Start  561: record_replay_subject-no-syscallbuf
          Start 1789: no_mask_timeslice-32-no-syscallbuf
```
That said, it's unfortunate that we didn't get the artifact for this run. I merged your other PR just now, but I'm not sure it would have helped. We may need to add another timeout to that script to start killing ctest before the buildkite timeout.

Keno:
And for my own future reference here's the quick two-liner to find the failing tests from the log:
```
lines = split(read(pipeline(`cat rr_build_61_build-and-test.log`, `grep -r "Test" -`), String), '\n')[2:end-1]
setdiff(1:2749, map(x->try; parse(Int, split(x, ' ', keepempty=false)[3][2:end-1]); catch; 0; end, lines))
```

DilumAluthge:
Which process do you want to kill? The `ctest` process?

DilumAluthge:
Interesting. This time, those two tests (`record_replay_subject-no-syscallbuf` and `no_mask_timeslice-32-no-syscallbuf`) did not time out! Instead, they both passed.

Keno:
We might have intermittent failures, but modulo that looks like we only have four tests (time four configurations) that are failing now and they all look relatively straightforward. I should be able to fix them tomorrow.

DilumAluthge:
Sounds good! In the meantime, this PR now only consists of the `verbose: true` line, which prints out the manifest that we are using. So it should be good to merge.

DilumAluthge:
> I should be able to fix them tomorrow.

Also, after you fix the tests, would you mind uploading a new `CTestCostData.txt` file? I imagine the cost data will be more accurate now.

PR #3070, opened 2022-01-25
Check for unmapped uid in cwd_inaccessible test

Keno:
Similar to #3069, unmapped UIDs in the container end up as
`(uint16_t)-2`. Treat that the same as the user not existing.


Keno:
@DilumAluthge we may want to clean up `/etc/group` and `/etc/passwd` in the sandbox, since only one UID/GID is available.

PR #3071, opened 2022-01-25
sysconf_onln: Skip test if sysfs is not mapped

Keno:


Keno:
`sysconf_onln` passed in the CI run.

PR #3072, opened 2022-01-25
setuid: Allow EINVAL error code

Keno:
Inside a user namespace, calling setgid with an unmapped `gid` will return
EINVAL, not EPERM. Accept that in case the environemnt we're running in
is a container.


Keno:
`setuid` passed in the CI run.

PR #3108, opened 2022-03-24
[DONOTMERGE] Try to debug cpu_lock code

Keno:
Manual investigation of CI runs shows that several tests
pin themselves to the same cores. Try to figure out what's
going on here.


Keno:
Ah. Looks like we never acquire the CPU lock for the bound CPU on replay.

PR #3109, opened 2022-03-24
Properly use cpu_lock for replay

Keno:
In rr, we have a mechanism to have several rr workers coordinate which
CPUs to pin to. However, because we have no freedom to choose the CPU
to pin to during replay, we were not marking those CPUs as in use, thus
allowing concurrent record calls to pin themselves to the same CPUs
(and even worse then causing their subsequent replay to also be bound
to the same CPU). Try to fix that by making sure to still acquire the
appropriate CPU lock in the replay path. There is still a small window
between the end of the record and the start of the replay where another
record could start on the same CPU, but let's assume that race isn't
hit for the moment.

### DIFF
--- a/src/Session.cc
+++ b/src/Session.cc
@@ -731,6 +731,9 @@ void Session::do_bind_cpu(TraceStream &trace) {
         FATAL() << "Can't bind to requested CPU " << cpu_index
                 << ", and CPUID faulting not available";
       }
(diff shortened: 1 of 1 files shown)


Keno:
I was watching this run and it looked good visually. Will merge and trigger another stampede to be sure.

PR #3110, opened 2022-03-24
CI: Use More parallelism

Keno:
With #3109 addressed, we should be able to use more parallelism on CI.
Let's try it.


Keno:
CI runs 5 mins faster with this. And I didn't see any excessive test times.

PR #3113, opened 2022-03-25
CI: before creating the tarball, delete any named pipes

DilumAluthge:


Keno:
Ref #3112 where we didn't get an artifact upload because tar couldn't tar up the named pipes that the test creates.

DilumAluthge:
@Keno Good to merge?

PR #3132, opened 2022-04-03
CI: more readable job names

DilumAluthge:
This PR allows us to have a more readable CI job name that includes spaces, while still having a key (not displayed directly to the end user) that adheres to Buildkite's restrictions.


DilumAluthge:
@Keno This PR allows us to have a more readable CI job name that includes spaces, while still having a key (not displayed directly to the end user) that adheres to Buildkite's restrictions.

DilumAluthge:
Bump @Keno

PR #3145, opened 2022-04-11
Remove hardcoded LSE atomic linker path

yuyichao:
1. This linker does not exist on ArchLinux ARM
2. From a brief search I believe all the load/store exclusive instructions within
   the normal ld-linux-aarch64.so.1 on ArchLinux ARM are guarded by feature tests
   so they shouldn't be used at runtime.


khuey:
I actually think we should just drop this entirely. Even on systems that have an LSE linker it could be in a different path on each distro. This really needs to be supplied by a distro maintainer.

yuyichao:
You added this in https://github.com/rr-debugger/rr/commit/713858c6b8c3d75bb75ae8d5557bb5a21c509e19 so does your normal ld contain direct use of the exclusive load/store instructions?

I'm certainly fine with dropping this. I assume by "supplied by distro maintainer" you mean to pass in the linker flag (e.g. via `LDFLAGS` env).

khuey:
> You added this in [713858c](https://github.com/rr-debugger/rr/commit/713858c6b8c3d75bb75ae8d5557bb5a21c509e19) so does your normal ld contain direct use of the exclusive load/store instructions?

Ubuntu's does, yes. I filed https://bugs.launchpad.net/ubuntu/+source/glibc/+bug/1933879

> I'm certainly fine with dropping this. I assume by "supplied by distro maintainer" you mean to pass in the linker flag (e.g. via `LDFLAGS` env).

Right.

yuyichao:
I changed this to get rid of the hard-coded path instead.

PR #3149, opened 2022-04-12
Fix sigcontext alignment on AArch64

yuyichao:
The actual structure contains saved vector registers and is therefore 16bytes aligned.
This causes the offset of uc_mcontext in ucontext to be wrong and in particular
the address where x7 is saved is wrong when we try to read it in the kernel bug workaround.

### DIFF
--- a/src/kernel_abi.h
+++ b/src/kernel_abi.h
@@ -2380,7 +2380,7 @@ struct ARM64Arch : public GenericArch<SupportedArch::aarch64, WordSize64Defs> {
     struct hw_bp dbg_regs[16];
   };
 
-  struct sigcontext {
+  struct __attribute((aligned(16))) sigcontext {
     __u64 fault_addr;
     user_pt_regs regs;
     // ISA extension state follows here


khuey:
Ha an excellent catch.

khuey:
Can you apply

```diff
diff --git a/src/kernel_abi.h b/src/kernel_abi.h
index 92503433..5352125c 100644
--- a/src/kernel_abi.h
+++ b/src/kernel_abi.h
@@ -36,11 +36,14 @@ template <SupportedArch a, typename system_type, typename rr_type>
 struct Verifier {
   // Optimistically say we are the same size.
   static const bool same_size = true;
+  // Optimistically say we are the same alignment.
+  static const bool same_align = true;
 };

 template <typename system_type, typename rr_type>
 struct Verifier<RR_NATIVE_ARCH, system_type, rr_type> {
   static const bool same_size = sizeof(system_type) == sizeof(rr_type);
+  static const bool same_align = alignof(system_type) == sizeof(rr_type);
 };

 template <typename T> struct Verifier<RR_NATIVE_ARCH, T, T> {
```

to your tree and see if that would have caught this? Does it catch anything else?

khuey:
mmm, this too

```diff
diff --git a/src/kernel_abi.cc b/src/kernel_abi.cc
index 50c9c85a..69b75ab9 100644
--- a/src/kernel_abi.cc
+++ b/src/kernel_abi.cc
@@ -65,7 +65,9 @@
 namespace rr {
 #define RR_VERIFY_TYPE_ARCH(arch_, system_type_, rr_type_)                     \
   static_assert(Verifier<arch_, system_type_, rr_type_>::same_size,            \
-                "type " #system_type_ " not correctly defined");
+                "type " #system_type_ " not correctly defined (size)");        \
+  static_assert(Verifier<arch_, system_type_, rr_type_>::same_align,           \
+                "type " #system_type_ " not correctly defined (align)");

 // For instances where the system type and the rr type are named differently.
 #define RR_VERIFY_TYPE_EXPLICIT(system_type_, rr_type_)                        \
```

This causes a ton of errors on x86 though, so maybe it's not what we want.

yuyichao:
I doubt it'll catch this one since not even the size of `sigcontext` and `ucontext` matches the actual definition. I assume this is intentionally left as such to not have to worry about matching things like sve states etc.

yuyichao:
In another word, it seems that the verify macro was intentionally left out for this..

khuey:
Ah, yes, that would be an issue for automatically checking it :)

PR #3151, opened 2022-04-13
Fix patch page end test segfault

yuyichao:
blr clobbers the link register causing the return from `do_call` to be completely messed up.

Similar to 8a8861a25777a444f75864a31e168ffd90799bd9, the test still doesn't pass but it should at least fail with the correct error now...

### DIFF
--- a/src/test/patch_page_end.c
+++ b/src/test/patch_page_end.c
@@ -37,7 +37,7 @@ static long do_call(uint8_t* p) {
 #elif defined(__aarch64__)
   register long x8 __asm__("x8") = SYS_getpid;
   register long x0 __asm__("x0") = 0;
-  __asm__ __volatile__("blr %2" : "=r"(x0) : "r"(x8), "r"(p));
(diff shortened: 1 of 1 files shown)


src/test/patch_page_end.c
```diff
@@ -37,7 +37,7 @@ static long do_call(uint8_t* p) {
 #elif defined(__aarch64__)
   register long x8 __asm__("x8") = SYS_getpid;
   register long x0 __asm__("x0") = 0;
-  __asm__ __volatile__("blr %2" : "=r"(x0) : "r"(x8), "r"(p));
+  __asm__ __volatile__("blr %2" : "=r"(x0) : "r"(x8), "r"(p) : "x30");
```
Keno, line 40: Does `lr` work here rather than `x30`? I can never remember which register is aliased where.

yuyichao: Unfortunately it doesn’t seem to. I tried lr, xlr and various capitalization. The disassembler never seem to do that (unlike 32 bit arm iirc) if that’s any indication.

I added a comment to hopefully makes it more clear…

PR #3158, opened 2022-04-14
Unconditionally print rr stack if something goes wrong

Keno:
In #3129, we're hitting some assertion, but the process is dead without
printing anything before we get a GDB backtrace. Hopefully this will shed
some light on the situation.


Keno:
@DilumAluthge Could I request a re-run of the 100 CI run torture test with this PR included to see if we get another instance of #3129 ?

PR #3183, opened 2022-04-22
Set state of new task as if just exiting from the syscall

yuyichao:
handle_ptrace_exit_event() has logic to detect this case and would otherwise record a ghost clone in the trace which would really confuse the replay.

I'm not 100% sure if this is the right fix though I'm 90% sure this is what's causing ~5-10 tests to fail randomly.


yuyichao:
Also, I'm not sure why I have not been able to reproduce it on x86 yet....

PR #3186, opened 2022-04-24
Fix CI by passing the correct `arch` parameter

DilumAluthge:


khuey:
I would merge this but it appears to still be broken :)

PR #3200, opened 2022-04-30
Fix syscall restart after signal on aarch64

yuyichao:
Unlike x86, the pc is already moved to before the syscall on aarch64 before the kernel inform us about it.

I have not really gone through all the code paths and confirm the state of the pc when it gets here but the change is consistent with the comment above https://github.com/rr-debugger/rr/blob/b8368dd96b7e25f4cc0b2faa6fd3eeadf6d0f22e/src/RecordSession.cc#L1614 as well as https://github.com/rr-debugger/rr/blob/b8368dd96b7e25f4cc0b2faa6fd3eeadf6d0f22e/src/ReplaySession.cc#L1786 .

I've also put in an assertion here to make sure that the instruction the ip points to is indeed the syscall instruction. There are about 40 tests that hits this path and all of them confirms the observation.

I suspect this don't usually show up because the instruction is fixed length and the instruction before the syscall can usually be reexecuted without any side-effects... (out of the ~400 instances of `svc 0` in my libc, only 2 are potentially problematic). This is a more severe issue with syscallbuf though since the syscalls on the rrpage are surrounded by `ret` that shouldn't be executed...

### DIFF
--- a/src/RecordSession.cc
+++ b/src/RecordSession.cc
@@ -1628,7 +1628,12 @@ bool RecordSession::signal_state_changed(RecordTask* t, StepState* step_state) {
               r.set_syscallno(syscall_number_for_restart_syscall(t->arch()));
               break;
           }
-          r.set_ip(r.ip().decrement_by_syscall_insn_length(t->arch()));
+          // On aarch64, the kernel modifies the registers before the signal stop.
+          // so we should not decrement the pc again or we'll rerun the instruction
+          // before the syscall.
+          // [1] https://github.com/torvalds/linux/blob/caffb99b6929f41a69edbb5aef3a359bf45f3315/arch/arm64/kernel/signal.c#L855-L862
+          if (t->arch() != aarch64)
+            r.set_ip(r.ip().decrement_by_syscall_insn_length(t->arch()));
           // Now that we've mucked with the registers, we can't switch tasks. That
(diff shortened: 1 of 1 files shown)


Keno:
LGTM on the aarch64 specific change, though I'm somewhat suspicious of this code path in general. I feel like the debugger would expect our semantics here to match the kernel, but as is, we've already applied the register changes on x86. I wonder if it wouldn't be better to do this with special handling in the replay path, but that's a separate consideration.

yuyichao:
Do you mean something needs to be different as a whole for this part? Since this is the recording part I assume what we need to do here to the callee shouldn't really depend on what the debugger want. The kernel and the hardware should be the only thing we need to deal with at this point. Do you mean that we need to do something differently in the record so that we'd do something different in replay?

yuyichao:
(Or are you talking about the aarch64 specific code path you added in [rr/src/ReplaySession.cc](https://github.com/rr-debugger/rr/blob/b8368dd96b7e25f4cc0b2faa6fd3eeadf6d0f22e/src/ReplaySession.cc#L1786) ?)

Keno:
Let's just do it like this for the moment. Anything else would be a larger refactor and would need new feature flags in the trace header, so it would need a strong motivation.

PR #3225, opened 2022-05-16
Implement PTRACE_(PEEK|POKE)(TEXT|DATA) on AArch64

yuyichao:
It appears that these are implemented and also used by rr itself.
This causes issue when recording a replay session without syscallbuf.


khuey:
Can you make the summary of the commit message something like "Move PTRACE_whatever from prepare_ptrace_legacy to prepare_ptrace"?

yuyichao:
done.

PR #3228, opened 2022-05-16
Allow override of prefix-map build flag

yuyichao:
By placing it before user supplied flags from the environment variables.


yuyichao:
I just realized that the prefix in the variable name and the variable value has nothing to do with each other.... Oh well, hopefully the comment at the variable initialization is clear enough...

CMakeLists.txt
```diff
@@ -69,18 +71,18 @@ configure_file(src/preload/rr_page.ld.in src/preload/rr_page.ld @ONLY)
 include(CheckCCompilerFlag)
 CHECK_C_COMPILER_FLAG("-fmacro-prefix-map=foo=bar" SUPPORTS_MACRO_PREFIX_MAP)
 if (SUPPORTS_MACRO_PREFIX_MAP)
-  set(FLAGS_COMMON "${FLAGS_COMMON} -fmacro-prefix-map=${CMAKE_SOURCE_DIR}/=")
+  set(PREFIX_FLAGS_COMMON "${PREFIX_FLAGS_COMMON} -fmacro-prefix-map=${CMAKE_SOURCE_DIR}/=")
```
Keno, line 74: How is this supposed to be used? In clang at least, `fmacro-prefix-map` is a hash table, with later entries overriding earlier ones. Doesn't that moot the usefulness of this option?

yuyichao: What do you mean by "usefulness of this option"? The point is that distribution should be able to override this. Without any flags passed in, this should behave exactly the same without this PR and as useful as it was.

It does seem that clang and gcc handles this differently. For GCC, it seems that any later flags override earlier ones but for clang, later, less-specific ones does not override earlier flags. To allow distro override to work correctly, it seems that I would need to just disable this if a similar flag is passed in.

Issue #2745, opened 2020-11-18
AArch64 syscallbuf support


khuey:
This is the big part of #1373 that is tractable and not done, so lets get it on file.

PR #3235, opened 2022-05-19
Implement syscallbuf on aarch64

### LINKED ISSUE #2603: Use of stack in syscall hooks
I was looking to doing the syscallbuf properly on aarch64 and I'm having trouble coming up with a sequence that does not require storing values on the original stack or leaving registers clobbered after the syscall. This is currently problematic because we explicitly test that syscalls work fine with a bad stack pointer. What do we want to assume about what we're allowed to clobber for the purposes of syscall patching? If we don't think we can clobber anything, the only other option I see would be to return using an extra sigreturn syscall or maybe to hide some state in the upper part of the SVE registers, which are documented to be clobbered during syscalls.

yuyichao:
Very WIP and request comment. I didn't realize that @Keno has already run into very much the same issue I did with the stack usage and there were already some discussion at #2603. I think I have come up with a solution that I'm fairly happy with. See a full write up of the design https://github.com/yuyichao/rr/wiki/Syscallbuf-on-AArch64. If it's ok I can move that into this wiki.

Fix #2603, #2745


yuyichao:
Sorry, meant to be a draft PR...

src/Monkeypatcher.cc
```diff
@@ -246,14 +259,193 @@ static remote_ptr<uint8_t> allocate_extended_jump(
...
+    ninst += 1;
+  }
+}
+
+/**
+ * Encode the following assembly.
+ *
+ *    cmp     x8, 1024
+ *    b.hi    .Lnosys
+ *    movk    x8, preload_thread_locals >> 16, lsl 16
```
Keno, line 285: Very clever

yuyichao:
And the current status, along with where I would expect the most feedback from.

* Currently there are about ~40 test failures with this. Roughly half of those are shared with master branch.
* The assembly code should have perfect unwind information. I manually checked that gdb is able to unwind every register values at every single instructions in the `_syscall_hook_trampoline_raw`. I didn't check `_syscall_hook_trampoline`, the inner function, as carefully. The unwinding is less tricky there and would mainly have issues if I have a typo.... I assume this will make @Keno happy. Do note that this require a fix in gdb to support unwinding pc and lr differently which was just merged in gdb master a few days ago.
* I have not seen anything from gdb that require me to do any dance with `__morestack`. I do think gdb devs are fairly open about fixing related issues so I don't currently see the necessity to do any of those tricks on aarch64 ATM.
* There are a few minor things that needs to be tweaked, e.g.

    * The assembly actually doesn't agree with the comment in `MonkeyPatcher.cc` (I'm actually not using the standard movz/movk sequence)
    * unwind info in assembly should use `cfi_rel_offset` to make it easier to understand (I only learnt about that after my latest refactor of the code.....)
    * inline assembly should use the `nf`, `nb` syntax to match with the style in the surrounding code and to avoid any label leaking into the final binary. I actually never knew about this syntax before....
    * I could also try to optimize some of the local memcpy/memset assembly implementations. It would just need a scalar and a vector loop basically. This would have a minor performance improvement for syscallbuf. It's too bad that aarch64 doesn't have anything like a string instruction that could be repurposed for a hardware memcpy/memset implementation.....

* And before anyone else mentions it, I think I'll need to make the huge buffer conditional on aarch64 and maybe also add some assertions to check it's offset. We could move it to elsewhere if there's a better place. I don't worry as much about the space wasting here, especially since we have another 60kB space to play with on aarch64 after https://github.com/rr-debugger/rr/pull/3146. I do worry a little bit about the additional memory copying during task switching but we could be a bit smarter in that regard. For a start, we never need to copy this part if we stopped in syscall, and we also never need to do anything about it unless x8 points into this area.

Other than these, I think there isn't fundamental change I'd like to make with the syscall buf implementation itself, which is why I'm posting this PR now (FWIW, this is roughly the 4th somewhat major rewrite of this implementation). I'd still like to understand most of the test failures better before having this merged to avoid as much regression as possible though so even if the syscallbuf implementation itself is perfect (doubt it...) I would not want to have this merged just yet....

khuey:
What tests are passing before but failing after this? Are they failing in any common way?

yuyichao:
The test result snapshot diffs is

```diff
+detach_state
 futex_restart_race
-futex_restart_race-no-syscallbuf
-async_signal_syscalls
-async_signal_syscalls2
-async_signal_syscalls_siginfo
-hooks
+intr_futex_wait_restart
+intr_read_no_restart
+intr_read_restart
+multiple_pending_signals_sequential
+orphan_process
+prctl_deathsig
+proc_fds
+proc_mem
+ptrace_breakpoint
+ptracer_death
+ptracer_death_multithread
+ptracer_death_multithread_peer
+sigprocmask_in_syscallbuf_sighandler
+syscall_bp
+syscallbuf_signal_blocking
+two_signals_with_mask
+async_kill_with_syscallbuf
+jit_proc_mem
+many_yields-no-syscallbuf
+mmap_replace_most_mappings
 morestack_unwind
-patch_page_end
-rseq_syscallbuf
 seccomp_blocks_rr
-vdso_clock_gettime_stack
-vdso_gettimeofday_stack
-vdso_time_stack
+step_thread
 watchpoint_at_sched
 watchpoint_at_sched-no-syscallbuf
 watchpoint_no_progress
 watchpoint_no_progress-no-syscallbuf
 desched_ticks
+fork_exec_info_thr
+get_thread_list
+record_replay
 replay_serve_files
 replay_serve_files-no-syscallbuf
 reverse_watchpoint
```

Do note that some of these are intermittent in either case.

I haven't found anything too obvious in common between the failures yet, though that's partially because I was mainly looking at one test and see how I can fix it. Someone with more experience with what these tests do would have a better idea I assume.

A lot of the issue seems to be somewhat related to signal handling. At least one of them manifests as a divergence during replay since somehow the replayed syscall isn't recognized as the traced syscall site and therefore didn't skip the register comparison.

Another test failure was related to something to do with single stepping (see the hack commit, which doesn't completely fix it either afaict).

yuyichao:
I'm pretty sure the `record_replay` is a timeout due to me running the tests in parallel (and that poor test was probably pinned to a cpu along with something else)

Keno:
> I'm pretty sure the `record_replay` is a timeout due to me running the tests in parallel (and that poor test was probably pinned to a cpu along with something else)

At least i theory, as long as you don't oversubscribe, rr processes will negotiate amongst themselves to avoid pinning themselves to the same core.

khuey:
Yeah my gut feeling looking at that list of tests is that there's something wrong with signals that hit while we're inside the syscallbuf.

yuyichao:
> At least i theory, as long as you don't oversubscribe, rr processes will negotiate amongst themselves to avoid pinning themselves to the same core.

I am actually oversubscribing a little... since quite a few of the test got stuck in the dignostic part (the reason I added the timeout env...). Could also be frequency throttling (the M1 actually does get warm when running tests for a few minutes) and maybe because it got pinned to a P core other than the E core. I might tweak the timeout numbers when I figured other things out....

> Yeah my gut feeling looking at that list of tests is that there's something wrong with signals that hit while we're inside the syscallbuf.

proc_fds, at least seems to be something else. AFAICT it's not signal related and I need to figure out why the tracee is seeing fd that it shouldn't be. Quite a few others are indeed signal related and it would really be nice if there's a summary of what's the expectation/invariance. I was able to do a line-by-line comparison with x86 with some luck but that's a bit slow....

rocallahan:
Your design in the wiki makes sense to me. The x8 trick is very clever. I think probably @Keno would be a better reviewer for this than me since he's worked on the syscallbuf code and also is more familiar with Aarch64.

src/Monkeypatcher.cc
```diff
@@ -737,6 +954,115 @@ bool Monkeypatcher::try_patch_syscall(RecordTask* t, bool entering_syscall) {
...
+    tried_to_patch_syscall_addresses.insert(ip);
+    return false;
+  }
+  // mov x8, 0xdc
+  if (inst[0] == 0xd2801b88) {
+    // Clone may either cause the new and the old process to share stack (vfork)
+    // or replacing the stack (pthread_create)
+    // and requires special handling on the caller.
+    // Our syscall hook cannot do that so this would have to be a raw syscall.
+    // We can handle this at runtime but if we know the call is definately
```
rocallahan, line 976: "definitely"

What's the point of this code? We do need to handle this at runtime somehow so we why is it worth also handling it here?

src/preload/overrides.c
```diff
@@ -179,44 +179,55 @@ void spurious_desched_syscall(struct syscall_info* info) {
...
+  int trash;
   asm volatile ("int $0x80; inc %0" : "=c"(trash) : "a"(SYS_sched_yield));
 #elif defined(__x86_64__)
+  // We have no syscall hook for `syscall` followed by `inc %ecx`
+  int trash;
   asm volatile ("syscall; inc %0" : "=c"(trash) : "a"(SYS_sched_yield));
+#elif defined(__aarch64__)
+  register long x8 __asm__("x8") = SYS_sched_yield;
+  // We explicitly blacklisted syscall that follows `mov x8, 0xdc`
+  // to avoid patching clone. Abuse that to prevent this from being patched.
```
rocallahan, line 203: I see that it's useful to have an instruction sequence that doesn't get patched, but if we only need that then it would be cleaner to check for a special instruction following `svc`.

src/Monkeypatcher.cc
yuyichao: I feel like it's still cleaner to handle it here and I assumed that if quite free to check since we are already reading tracee memory anyway. I originally thought that I have to make the check in the stub and make the syscall there which would mean that the clone will be issues without a usable backtrace but since I am actually making it in librrpreload.so that's actually not a problem anymore.

This does match what x86 does better though. The fork/vfork/clone actually never get patched on x86 AFAICT. This is the main reason I decided to keep it even after I have a runtime fix for the general case...

src/preload/overrides.c
yuyichao: This actually comes **after** the pattern matching. I was going to add a check for a strange instruction following svc but realize that I could abuse the currently one so I didn't bother adding anything......

src/preload/syscall_hook.S
```diff
@@ -572,6 +572,440 @@ SYSCALLHOOK_START(_syscall_hook_trampoline_49_89_ca)
...
+        stp     q0, q1, [sp, -688]!
+        .cfi_def_cfa_offset 688
+        .cfi_rel_offset q0, 0
+        .cfi_rel_offset q1, 0 + 16
+        stp     x29, x30, [sp, 32]
+        .cfi_rel_offset x29, 32
+        .cfi_rel_offset x30, 32 + 8
+        /* Build a |struct syscall_info| on the stack by pushing the arguments
+           and syscall number. */
+        stp     x8, x0, [sp, 48]
```
Keno, line 606: Factor out this pattern (with the cfis) into a `#define`, e.g. `STORE_PAIRX(x8, x0, 48)` to avoid any accidental future copy/paste bugs if we ever need to adjust it?

src/Monkeypatcher.cc
rocallahan: Why aren't fork/vfork/clone patched on x86? We just don't have a syscall hook for the code pattern they use?

IMHO having two ways to handle clone is not cleaner than having only one way to handle it.

src/preload/syscall_hook.S
```diff
@@ -572,6 +572,440 @@ SYSCALLHOOK_START(_syscall_hook_trampoline_49_89_ca)
...
+        movk    x29, #:abs_g0_nc:alt_stack_nesting_level
+        ldr     w30, [x29]
+        sub     w30, w30, 1
+        str     w30, [x29]
+
+        ldp     x29, x30, [sp, 32]
+        .cfi_restore x29
+        .cfi_restore x30
+        ldr     x8, [sp, 48]
+        .cfi_restore x8
```
Keno, line 697: IIRC the default CFI for x0-x15 is cfi_undefined, so these should probably be `cfi_same_value`.

src/Monkeypatcher.cc
yuyichao: AFAICT the code pattern for those aren't handled (actually I'm less sure about fork but I'm pretty sure about clone and vfork).

I was trying to see what x86 does and I realized that the syscall site didn't get patched.... On my laptop (archlinux x86_64) `clone` has `syscall` followed by `test %rax,%rax` and conditional branch whereas `vfork` has the syscall followed by `push %rdi`. None of these seems to be covered by the patterns.

I guess clean isn't the right word. I'm mainly saying that for the most common case (I assume few code use syscall to call clone) if we know the call is almost certainly something that will skip syscallbuf, we can just avoid the patch to keep perturbation on the callee to be minimum. Given that this path isn't tested on x86, I was also worrying a little if there's any other side effects that I haven't considered.....

src/preload/syscall_hook.S
yuyichao: Sure that works. I think gdb at least has a comment essentially saying that it'll treat absence of unwind info as same_value since gcc has/had a habbit of doing that..... I guess not every unwinder has that hack....

yuyichao: I assume to be consistent I would also need to prefix other functions with a bunch of `.cfi_same_value`s though... I think this would even include most of the ASIMD registers.

src/preload/syscall_hook.S
```diff
@@ -572,6 +572,440 @@ SYSCALLHOOK_START(_syscall_hook_trampoline_49_89_ca)
...
+        .cfi_escape     0x10, /* DW_CFA_expression */                   \
+                        0x20, /* pc */                                  \
+                        0x06, /* length 6 */                            \
+                        0x0c, /* DW_OP_const4u */                       \
+                        stub_scratch_1 & 0xFF,                          \
+                        (stub_scratch_1 & (0xFF <<  0x8)) >>  0x8,      \
+                        (stub_scratch_1 & (0xFF << 0x10)) >> 0x10,      \
+                        (stub_scratch_1 & (0xFF << 0x18)) >> 0x18,      \
+                        0x06, /* DW_OP_deref */
+        // Move the saving area to the start of scratch_2
```
Keno, line 830: I was a little bit confused by this comment at first. Maybe something like
```suggestion
        // Move the register stash region from `x8` (i.e. `stub_scratch_2 + original_x8`) to the start of `stub_scratch_2`
```

src/preload/syscall_hook.S
```diff
@@ -572,6 +572,440 @@ SYSCALLHOOK_START(_syscall_hook_trampoline_49_89_ca)
...
+                        0x00
+        .cfi_escape     0x16, /* DW_CFA_val_expression */                       \
+                        0x08, /* x8 */                                          \
+                        0x06, /* length 6 */                                    \
+                        0x78, /* DW_OP_breg8 */                                 \
+                        ((-preload_thread_locals) & 0x7F) | 0x80,               \
+                        ((-preload_thread_locals) & (0x7F << 7)) >> 7 | 0x80,   \
+                        ((-preload_thread_locals) & (0x7F << 14)) >> 14 | 0x80, \
+                        ((-preload_thread_locals) & (0x7F << 21)) >> 21 | 0x80, \
+                        ((-preload_thread_locals) & (0x7F << 28)) >> 28 | 0x70
```
Keno, line 816: Why is the `0x70` here required?

src/preload/syscall_hook.S
```diff
@@ -572,6 +572,440 @@ SYSCALLHOOK_START(_syscall_hook_trampoline_49_89_ca)
...
+        str     x30, [x15, stub_scratch_2 - preload_thread_locals]
+        ldr     x30, [x8, stub_scratch_2 - preload_thread_locals + 8]
+        str     x30, [x15, stub_scratch_2 - preload_thread_locals + 8]
+        // Restore x8
+        movk    x8, 0, lsl 16
+        .cfi_restore x8
+        .cfi_escape     0x10, /* DW_CFA_expression */                   \
+                        0x0f, /* x15 */                                 \
+                        0x05, /* length 5 */                            \
+                        0x0c, /* DW_OP_const4u */                       \
```
Keno, line 842: I think at this point this pattern is used often enough to warrant a macro to encode this.

src/preload/syscall_hook.S
```diff
@@ -572,6 +572,440 @@ SYSCALLHOOK_START(_syscall_hook_trampoline_49_89_ca)
...
+
+        b.ne    .Lnest_syscall_hook_trampoline_raw
+        ldr     x30, [x15, syscallbuf_stub_alt_stack - preload_thread_locals]
+        sub     x30, x30, 48
+        b       .Lstackset_syscall_hook_trampoline_raw
+.Lnest_syscall_hook_trampoline_raw:
+        sub     x30, sp, 48
+.Lstackset_syscall_hook_trampoline_raw:
+        // Now x30 points to the new stack with 48 bytes of space allocated
+        .cfi_def_cfa x30, 48
```
Keno, line 873: This looks like the CFA changes halfway through the function. That will confuse the unwinder. You need to make sure that the CFA has the same at every instruction in the function.

src/preload/syscallbuf.c
```diff
@@ -194,6 +194,18 @@ static void local_memcpy(void* dest, const void* source, int n) {
                        : "+S"(source), "+D"(dest), "+c"(n)
                        :
                        : "cc", "memory");
+#elif defined(__aarch64__)
+  long c;
+  __asm__ __volatile__("cbz %3, 2f\n"
```
Keno, line 199: Do we need something faster here and link in the standard assembly memcpy routine?

Keno:
I've left some specific comments, but overall this looks good to me.

src/preload/syscall_hook.S
Keno: Yeah, that does sound annoying. I guess that's supposed to be done in the CIE, but it doesn't really let you encode those directly.

src/preload/syscall_hook.S
yuyichao: I don't remember how generic this encoding was but I'm pretty sure this is to get the sign of the LEB128 right...

src/preload/syscall_hook.S
yuyichao: I checked that GDB is fine with this as least when I single instruction through the function.
I am wondering what the CFA should be though... The aadwarf64 defined it as the SP before the function call. I can do that fairly easily (need to figure out how to encode it.) but then I guess I just can't use that to save/restore any registers since nothing will be stored relative to that. Alternatively, since at least gdb seems fairly happy with my explicit sp restoring unwind info, I could define CFA as the actual top of the stack frame in this function. This is what I'm doing here, but it's near impossible to encode the CFA before reaching here (unless I essentially write this stack pointer bumping logic again in dwarf expression....). Which way do you think is better?

src/preload/syscall_hook.S
yuyichao: You just mean the

```asm
stp xm, xn, [sp, a]
.cfi_rel_offset xm, a
.cfi_rel_offset xn, a + 8
```

pattern right? Yeah, sure. Now that I know the patterns that I used a lot I'll convert some of those into macros. I would maybe also define two different ones for x and q and pass in only the register numbers since otherwise you could accidentally do a `STORE_PAIRX(q0, q1, 20)` and get the wrong offset for q1...

src/preload/syscall_hook.S
yuyichao: Yeah I think these are the sign bits I need to get a 35bits negative number from a 32bit negative number. I might be able to define a constant like `-preload_thread_locals | (7 << 32)` and use that in all the expressions. I wasn't sure if I need to worry about integer type though (that constant must not be 32bit) which is why I didn't do it that way...

src/preload/syscall_hook.S
```diff
@@ -572,6 +572,440 @@ SYSCALLHOOK_START(_syscall_hook_trampoline_49_89_ca)
...
+        .cfi_rel_offset q24, 560
+        .cfi_rel_offset q25, 560 + 16
+        stp     q26, q27, [sp, 592]
+        .cfi_rel_offset q26, 592
+        .cfi_rel_offset q27, 592 + 16
+        stp     q28, q29, [sp, 624]
+        .cfi_rel_offset q26, 624
+        .cfi_rel_offset q27, 624 + 16
+        stp     q30, q31, [sp, 656]
+        .cfi_rel_offset q26, 656
```
yuyichao, line 682: And point taken regarding copy-n-paste mistake................... @Keno

src/preload/syscall_hook.S
Keno: Use the sp when entering the function as the CFA. Yes, it'll make encoding the saved registers more annoying, but I think it's the most correct. Otherwise you'd have to replicate the CFA computation logic in DWARF at the start of the function, which is ... Unfun.

src/preload/syscall_hook.S
Keno: Ok. That's reasonable, but the please
A) adjust the mask for only the bits you pick up to make it clear and
B) add a comment

Keno (approved):

yuyichao:
As @keno probably noticed, I’ve addressed most of the review comments. I’ve also made some progress on debugging the signal issue (which I think has much to do with the value of pc after an interrupted syscall) . This pr still contains hacks and regressions though so it’s not ready to be merged just yet…

yuyichao:
The direct reason for the signal-related test failure appears to be caused by pc pointing to the beginning of the syscall instruction on aarch64 when it generates the signal stop. This throws off the code checking where we are and probably violated some invariance....

I have a very hacky patch that fixes most of the new test failures but I assume it might be better to do the refactoring @Keno wanted in https://github.com/rr-debugger/rr/pull/3200 ?

```diff
diff --git a/src/Monkeypatcher.cc b/src/Monkeypatcher.cc
index 285f80d5..b774a8d9 100644
--- a/src/Monkeypatcher.cc
+++ b/src/Monkeypatcher.cc
@@ -1028,7 +1028,7 @@ bool Monkeypatcher::try_patch_syscall(RecordTask* t, bool entering_syscall) {
     // forwards, that the syscall occurs deep in the preload library instead.
     return false;
   }
-  if (t->is_in_traced_syscall()) {
+  if (t->is_in_traced_syscall(t->ip())) {
     // Never try to patch the traced-syscall in our preload library!
     return false;
   }
diff --git a/src/Task.h b/src/Task.h
index 5e110d87..2b349b7b 100644
--- a/src/Task.h
+++ b/src/Task.h
@@ -325,17 +325,17 @@ public:
    * is implied by this. Note that once we've entered the traced syscall,
    * ip() is immediately after the syscall instruction.
    */
-  bool is_in_traced_syscall() {
-    return ip() ==
+  bool is_in_traced_syscall(remote_code_ptr ip) {
+    return ip ==
                as->traced_syscall_ip().increment_by_syscall_insn_length(
                    arch()) ||
-           ip() ==
+           ip ==
                as->privileged_traced_syscall_ip()
                    .increment_by_syscall_insn_length(arch());
   }
-  bool is_at_traced_syscall_entry() {
-    return ip() == as->traced_syscall_ip() ||
-           ip() == as->privileged_traced_syscall_ip();
+  bool is_at_traced_syscall_entry(remote_code_ptr ip) {
+    return ip == as->traced_syscall_ip() ||
+           ip == as->privileged_traced_syscall_ip();
   }
 
   /**
@@ -345,10 +345,13 @@ public:
    * entered the traced syscall, ip() is immediately after the syscall
    * instruction.
    */
-  bool is_in_untraced_syscall() {
-    auto t = AddressSpace::rr_page_syscall_from_exit_point(arch(), ip());
+  bool is_in_untraced_syscall(remote_code_ptr ip) {
+    auto t = AddressSpace::rr_page_syscall_from_exit_point(arch(), ip);
     return t && t->traced == AddressSpace::UNTRACED;
   }
+  bool is_in_untraced_syscall() {
+    return is_in_untraced_syscall(ip());
+  }
 
   bool is_in_rr_page() {
     auto p = ip().to_data_ptr<void>();
diff --git a/src/record_signal.cc b/src/record_signal.cc
index 17360554..156bb59e 100644
--- a/src/record_signal.cc
+++ b/src/record_signal.cc
@@ -304,7 +304,7 @@ bool handle_syscallbuf_breakpoint(RecordTask* t) {
   r.set_ip(r.ip().undo_executed_bkpt(t->arch()));
   t->set_regs(r);
 
-  if (t->is_at_traced_syscall_entry()) {
+  if (t->is_at_traced_syscall_entry(t->ip())) {
     // We will automatically dispatch stashed signals now since this is an
     // allowed place to dispatch signals.
     LOG(debug) << "Allowing signal dispatch at traced-syscall breakpoint";
@@ -575,7 +576,12 @@ static bool is_safe_to_deliver_signal(RecordTask* t, siginfo_t* si) {
     return true;
   }

-  if (t->is_in_traced_syscall()) {
+  auto ip = t->ip();
+  if (t->arch() == aarch64 && t->ev().type() == EV_SYSCALL_INTERRUPTION) {
+    ip = ip.increment_by_syscall_insn_length(t->arch());
+  }
+
+  if (t->is_in_traced_syscall(ip)) {
     LOG(debug) << "Safe to deliver signal at " << t->ip()
                << " because in traced syscall";
     return true;
@@ -589,14 +595,14 @@ static bool is_safe_to_deliver_signal(RecordTask* t, siginfo_t* si) {
   // that will be too late for syscallbuf to notice.
   // It's OK to delay signal delivery until after rrcall_notify_syscall_hook_exit
   // anyway.
-  if (t->is_at_traced_syscall_entry() &&
+  if (t->is_at_traced_syscall_entry(ip) &&
       !is_rrcall_notify_syscall_hook_exit_syscall(t->regs().syscallno(), t->arch())) {
     LOG(debug) << "Safe to deliver signal at " << t->ip()
                << " because at entry to traced syscall";
     return true;
   }

-  if (t->is_in_untraced_syscall() && t->desched_rec()) {
+  if (t->is_in_untraced_syscall(ip) && t->desched_rec()) {
     // Untraced syscalls always use the architecture of the process
     LOG(debug) << "Safe to deliver signal at " << t->ip()
                << " because tracee interrupted by desched of "
@@ -606,7 +612,7 @@ static bool is_safe_to_deliver_signal(RecordTask* t, siginfo_t* si) {
     return true;
   }

-  if (t->is_in_untraced_syscall() && si->si_signo == SIGSYS &&
+  if (t->is_in_untraced_syscall(ip) && si->si_signo == SIGSYS &&
       si->si_code == SYS_SECCOMP) {
     LOG(debug) << "Safe to deliver signal at " << t->ip()
                << " because signal is seccomp trap.";
@@ -625,7 +631,7 @@ static bool is_safe_to_deliver_signal(RecordTask* t, siginfo_t* si) {

     // A signal (e.g. seccomp SIGSYS) interrupted a untraced syscall in a
     // non-restartable way. Defer it until SYS_rrcall_notify_syscall_hook_exit.
-    if (t->is_in_untraced_syscall()) {
+    if (t->is_in_untraced_syscall(ip)) {
       // Our emulation of SYS_rrcall_notify_syscall_hook_exit clears this flag.
       t->write_mem(
           REMOTE_PTR_FIELD(t->syscallbuf_child, notify_on_syscall_hook_exit),
```

yuyichao:
The latest test run on a M1 has a single failure left (maybe there are a few more intermittent one but I've fixed a few of those as well) which is `reverse_step_signal-no-syscallbuf` and is shared with the master branch. (somehow gdb did one more reverse single step after reversing out of the signal handler in reverse-finish).

The last few commits are the prototype for the fixes and I'll submit them separately. Some of these command are going to be more controversial than others.....

src/RecordSession.cc
```diff
@@ -171,7 +171,11 @@ static bool handle_ptrace_exit_event(RecordTask* t) {
       // has been updated to reflect syscall entry. If we record a SCHED in
       // that state replay of the SCHED will fail. So detect that state and fix
       // it up.
-      if (looks_like_syscall_entry(t)) {
+      // If we got killed in an untraced syscall,
+      // there's no way on aarch64 that we can distinguish between
```
Keno, line 175: There's PTRACE_GET_SYSCALL_INFO. Also that is technically what x7 tells you on aarch64, but that's terrible of course.

yuyichao: I did see that after I wrote this fix. I wasn’t sure if that’s usable here both because we are not at a syscall stop and the tracee might be dead already…. The manual for that says “Retrieve information about the system call that caused the stop.” But we got here due to an exit…

it seems to me that for untraced syscall not recording them shouldn’t matter here since it should be indistinguishable from the case where we got killed just before or after syscall (I.e. the period between starting the buffered syscall till committing it should have no side effects that needs to be replayed on exit.) I assume the case this is handling is for an unbuffered syscall.

yuyichao:
Finally just had the very first completely successful test run locally on this branch :tada: 

```
100% tests passed, 0 tests failed out of 1317
```

There are still a few intermittent test failures left though and I believe all of them are related to predicting the pc accurately when we got a signal.... A few other hacks on this branch also still needs to be tweaked...

khuey:
Congratulations. You've done fantastic work on this.

yuyichao:
The `vm_readv_writev` failure is reproducible on the host machine and also without running under rr so I assume it's some kernel behavior that's different.

The `early_error` assertion doesn't reproduce in the host.

yuyichao:
So the `gdb_qpasssignals` test that times out on the CI does actually finish. It just takes 1.5 hours to do that, not sure what makes it that slow yet.....

rocallahan:
I guess `qPassSignals` is not working for some reason, then. It tries to reverse-execute through a large number of signals, which will be incredibly slow if the signals aren't being filtered by rr.

yuyichao:
Doesn't seem to be as simple it seems?

This works fine without syscallbuf. On the test machine where it doesn't works with syscallbuf, the reversing is going through the `GdbConnection::notify_stop` code path whereas on M1 with syscallbuf it's **not**. It is still going through the `GdbConnection::is_pass_signal` code path though. I compiled a gdb master on the test machine and at least that doesn't seem to be the issue....

yuyichao:
OK, there are still minor things to tweak/fix but I think this is ready now.

All the tests now passes quite reliably on my M1 and RB5. I've seen one or two tests that fails maybe around 1% of the time and I know at least one cause of it (see below). There are 3 remaining test failures on the CI, one of which (qpasssignals) I could reproduce and debug. The other one is still unclear... (maybe it's the special glibc)

I've recorded firefox but the replay failed with tick difference and I suspect it's because I still have llsc on my system.

Main changes from the initial/previous version:

* Return address is stored at the end of the stub now. I originally did this since I thought this is necessary for return address signing but I don't think that's the case. This would maybe make the return address predictor more happy but I've not seen an effect on M1 and we are most likely doing at least a syscall in between so I don't think it'll matter much...
* Added BTI support for the syscall hook. If someone want to map librrpreload as hardened library it should be fine now...
* Unlike x86, where we return straight to user code from `syscall_hook.S`, we need to go through the stub on aarch64. The code in the stub is also still using the scratch memory and we need to avoid interrupting it with a signal before it's down with the memory. This is now handled by adding a breakpoint on the stub exit if we were in a stub when we got the signal. We could potentially simply manually single step it but the current way fits the current code better.

Known issues/missing optimization/other TODOs:

* Avoid copying stub_scratch_2 if we don't need to
* Maybe tweak manual stub return handling (we may not have to put the return address in stub scratch 1 and we may not need the stub return breakpoint)
* Maintaining the condition code across syscalls. We can do this conditionally for syscalls by checking if there's any instructions that overwrite them. This may be more useful when we can trap/patch `mrs` instructions for getting the time though in that case we'll have an easy register to overwrite and the stub code would be much simpler...
* Known test failures

    * `cont_race` happens occationally when we try to read the memory sideeffect of a syscall and the tracee exited unexpectedly.
    * `many_yields-no-syscallbuf` just saw an assertion in this test over 30 test runs this morning.
    * `early_error*` on CI
    * `gdb_qpasssignals` on CI

Keno:
I would suggest merging this right away. We can get to the bottom of the remaining failures afterwards.

yuyichao:
I copied the wiki page here https://github.com/rr-debugger/rr/wiki/Syscallbuf-on-AArch64 . May not stay in sync but should still be useful to understand some of the weird assembly dance....

PR #3236, opened 2022-05-19
Add an assertion to make sure we never try to patch syscall from rrpage

yuyichao:
This is cherry-picked from #3235 and this is actually more of a request for clarification than useful patch.

I do not see a check for this elsewhere, the closest is the `t->is_in_traced_syscall()` above but that only cover 2 of the many syscall sites in the rr page. On AArch64 with syscallbuf and a broken syscall restart detection (i.e. https://github.com/yuyichao/rr/commit/5f6e008ec7f671dc4acf021cbc233a1b00ca27e9) I am seeing code reaching here with a rrpage address and fixing the syscall restart logic fixes it. I did the fix, which I still need to convince myself whether it's correct/complete or not, only because I see the x86 version diverges with the aarch64 from here but I don't really know what is the invariant that should prevent this from happening.

That said, I do think an explicit check here, either in the branch condition or as an assertion, would be useful since currently if we actually enter this function with a rr page address, the pattern mismatch would still prevent it from being patched on x86 but not on aarch64. An explicit check here would make the behavior more consistent between archs.

### DIFF
--- a/src/Monkeypatcher.cc
+++ b/src/Monkeypatcher.cc
@@ -582,6 +582,12 @@ bool Monkeypatcher::try_patch_syscall(RecordTask* t, bool entering_syscall) {
 
   Registers r = t->regs();
   remote_code_ptr ip = r.ip();
+  // We should not get here for untraced syscalls or anything else from the rr page.
+  // These should be normally prevented by our seccomp filter
+  // and in the case of syscalls interrupted by signals,
+  // the check for the syscall restart should prevent us from reaching here.
+  DEBUG_ASSERT(ip.to_data_ptr<void>() < AddressSpace::rr_page_start() ||
+               ip.to_data_ptr<void>() >= AddressSpace::rr_page_end());
   if (tried_to_patch_syscall_addresses.count(ip)) {
     return false;
   }


yuyichao:
(In another word, if someone can clarify what is the guarantee/assumption here, I'd like to put in a comment about it together with an assertion or an additional branch condition)

khuey:
`is_in_traced_syscall()` should cover all of the cases that can end up in the rr supervisor during recording. The remaining syscalls are either untraced (so they should go directly to the kernel thanks to our seccomp filter) or replay only (so the logic in RecordSession that calls try_patch_syscall is irrelevant).

yuyichao:
I was initially confused as well and thought that the seccomp isn't working. However, I saw that if a syscall is interrupted by signal, in some cases, we would see it stopped in rr at one of the other syscall addresses. It seems that the syscall restart check is the only thing that covers that case and prevents it from reaching here.

So basically my question is that whether the understanding above is correct and expected.

khuey:
I think that's correct, yes. The `try_patch_syscall` invocation in `RecordSession::process_syscall_entry` is gated on `!maybe_restart_syscall()` which I think should take care of any cases where an unbuffered syscall is interrupted by a signal.

Keno:
> `RecordSession::process_syscall_entry` is gated on `!maybe_restart_syscall()`

Just drive by noting that syscall restart looks different on aarch64, so perhaps that logic is broken, but I'm sure @yuyichao is well aware of that

yuyichao:
> Just drive by noting that syscall restart looks different on aarch64, so perhaps that logic is broken, but I'm sure @yuyichao is well aware of that

I have suspected that and that’s one of the reasons I’m not so confident about the fix commit. Any pointer where I can find a description on how it behaves? I was going to try some trace myself if I couldn’t find anything……

yuyichao:
Added a comment with the assumption used here.

PR #3239, opened 2022-05-21
Print orig_arg1 value on aarch64

yuyichao:
On x86 this is exactly the same as arg1 but this is not the case on aarch64

I swear that failing to update orig_arg1 accounts for at least half of the test failures that I've seen so far on aarch64......


khuey:
> I swear that failing to update orig_arg1 accounts for at least half of the test failures that I've seen so far on aarch64......

That does not surprise me.

PR #3241, opened 2022-05-21
Fix more breakpoint handling on aarch64

yuyichao:
Fix code that assumes the pc points to after the breakpoint instruction
after hitting the breakpoint.

### DIFF
--- a/src/remote_code_ptr.h
+++ b/src/remote_code_ptr.h
@@ -52,12 +52,25 @@ public:
       case x86:
       case x86_64:
         return decrement_by_bkpt_insn_length_private(arch);
-      default:
-        DEBUG_ASSERT(0 && "Unknown architecture");
-        RR_FALLTHROUGH;
       case aarch64:
         // On aarch64 executed breakpoint instructions do not increment the pc
         return *this;
+      default:
+        DEBUG_ASSERT(0 && "Unknown architecture");
+        __builtin_unreachable();
+    }
+  }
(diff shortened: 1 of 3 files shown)


src/remote_code_ptr.h
```diff
@@ -60,6 +60,19 @@ class remote_code_ptr {
         return *this;
     }
   }
+  remote_code_ptr continue_executed_bkpt(SupportedArch arch) {
```
khuey, line 63: This I think needs a better name. I have no idea what it's supposed to do.

yuyichao: I'm very bad at coming up with names so I'm very open to suggestions....

I was initially going to call it "redo_****" but that feels really like making it redo the breakpoint, which is exactly the opposite.... so...

Keno: advance_past_executed_bkpt?

khuey: that sounds good to me.

Also put the aarch64 case above default ...

yuyichao: Sure,

> Also put the aarch64 case above default ...

What about the undo function? I've got confused by that order before and I just copied it here.....

khuey: Yeah you can fix that too.

yuyichao: Done. It seems that the ordering could be for the `RR_FALLTHROUGH`. I changed that to a unreachable instead as already used elsewhere.

PR #3245, opened 2022-06-01
Support bionic's PR_SET_VMA.

khuey:
This is used to communicate additional information to the kernel about anonymous VMAs and prevent merging of VMAs with different names. It does not require any special handling by rr.

I'm requesting review on this only because I wonder if you think I should wrap this in #ifdef __BIONIC__?


Keno:
The prctl is in mainline Linux as of 5.17: https://github.com/torvalds/linux/blob/143a6252e1b8ab424b4b293512a97cca7295c182/include/uapi/linux/prctl.h#L284-L285

khuey:
Well that answers my question then. Thanks!

src/record_syscall.cc
```diff
@@ -4598,8 +4598,24 @@ static Switchable rec_prepare_syscall_arch(RecordTask* t,
               syscall_state.expect_errno = EINVAL;
               break;
           }
+          break;
+        }
+
+        case PR_SET_VMA: {
+          switch (regs.arg2()) {
+            case PR_SET_VMA_ANON_NAME:
+              // PR_SET_VMA_ANON_NAME is used by Android to communicate
```
Keno, line 4607: nit: The Android reference here seems somewhat misleading since this is in mainline now.

PR #3258, opened 2022-06-06
Fix hardware breakpoint handling and test on AArch64

yuyichao:
See commit message for detail.


yuyichao:
Also, interestingly, gdb doesn't support hardware breakpoint directly on aarch64 but rr does. That initially led me into the wrong direction before I checked that it indeed works on non-syscall instructions......

PR #3273, opened 2022-06-17
Use PT_INTERP to find the dynamic linker/program interpreter and record that information in the trace, for later use by --serve-files.

khuey:
bionic doesn't use the glibc linker and it doesn't follow the same naming convention. This seemed like a good opportunity to bite the bullet and use PT_INTERP here. This is complicated slightly by the fact that the GdbServer can also be invoked during _recording_ by the emergency debugger if rr asserts. In practice the emergency debugger won't use --serve-files so it shouldn't matter, but it's easy enough to put the interpreter name/base address on AddressSpace, which is the natural place to smuggle this information from the task exec event to GdbServer, in both recording and replay.

### DIFF
--- a/src/AddressSpace.h
+++ b/src/AddressSpace.h
@@ -359,6 +359,12 @@ public:
    */
   const std::string& exe_image() const { return exe; }
 
+  const std::string& interp_name() const { return interp_name_; }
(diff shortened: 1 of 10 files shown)


Keno (approved):
Seems good to me.

PR #3276, opened 2022-06-18
Fix PMU count check

yuyichao:
Allow **single** PMU type to use RAW event, not **zero** PMU types...

Typo from #3265

This is related to the current CI failure on aarch64 since it broke the PMU detection on neoverse-n1 when there's no sysfs.

@DilumAluthge @Keno It seems that sysfs isn't available/accessible on CI. I added a command that print out the warning and it seems that the reading of sysfs for CPUID and PMU types both failed. https://buildkite.com/julialang/rr/builds/674#0181787e-8097-44f0-ba1d-349522b98b87/395-2280

### DIFF
--- a/src/PerfCounters_aarch64.h
+++ b/src/PerfCounters_aarch64.h
@@ -365,7 +365,7 @@ static void post_init_pmu_uarchs(std::vector<PmuConfig> &pmu_uarchs)
       pmu_uarchs.resize(1);
     }
   }
-  if (pmu_uarchs.size() != 0 && pmu_type_failed) {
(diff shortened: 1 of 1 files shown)


DilumAluthge:
> @DilumAluthge @Keno It seems that sysfs isn't available/accessible on CI.

I'll follow up with Elliot on Slack about getting  https://github.com/staticfloat/Sandbox.jl/releases/tag/v1.4.0 deployed to our agents.

yuyichao:
Ah ok, that explains why there still seems to be errors about pidof missing as well.

PR #3277, opened 2022-06-19
CI: update to the latest rootfs images

DilumAluthge:


DilumAluthge:
@yuyichao This should give us `pidof`.

DilumAluthge:
On aarch64, we are now down to:

```
98% tests passed, 28 tests failed out of 1319
```

yuyichao:
Somehow pidof is still missing.... https://buildkite.com/julialang/rr/builds/680#01817dc7-0237-4937-9c36-e4b7db5e947f/6-938 maybe I got the wrong package name? I don't have an actual system to test but my source was https://packages.debian.org/bullseye/arm64/sysvinit-utils/filelist ...

PR #3282, opened 2022-06-20
Add part numbers for M1 Pro

tbodt:
They are different from M1 for some reason. Everything else seems to work fine.


tbodt:
@yuyichao

yuyichao:
Given that the previous X versions (e.g. A13X vs A13) also have different midr values I’m not surprised… At least apple knows how to count… unlike ARM. I don’t have hardware to check but I assume you got the right values…

tbodt:
After trying to use it more seriously I did end up with a crash when replaying a trace. I can't tell if this is a bug in arm support more generally or specific to M1 max.

PR #3283, opened 2022-06-21
Wait for detached nested record to finish before replaying it

yuyichao:
By waiting for the incomplete file to disappear

This is more likely to happen on asymmetric cores with very different performance.

On my M1 MBA, this happens ~20 times for `nested_detach` in 80 runs of the full test suite. This can also be easily reproduced by adding a sleep in `simple.c` or in `TraceStream.cc` before `incomplete` is renamed to `version`.

### DIFF
--- a/src/test/nested_detach.run
+++ b/src/test/nested_detach.run
@@ -8,5 +8,6 @@ check_record
 # Replay inner
 cd inner
 workdir=$PWD
+wait_for_complete
 replay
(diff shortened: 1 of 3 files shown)


yuyichao:
Actually, I just realized that this probability (1/4) agrees remarkably well with the probability that the the parent is assigned to a P core and the child is assigned to a E core.

PR #3288, opened 2022-06-25
Handle unexpected tracee exit in syscallbuf initialization

yuyichao:
This actually happens the most often when I run the `cont_race` test in a racy way (multiple run sharing the same `fifo`)....


yuyichao:
Forgot to delete the else case a few lines down… I can do hat when I’m home …

PR #3307, opened 2022-07-02
Try fixing early_error test

Keno:
On x86, the syscall we expect to veto here is the prctl that sets
PR_SET_TSC (and thus happens before the SIGSTOP that indicates
completion of the intial setup code). However, on aarch64 this
prctl doesn't exist. Depending on ordering of events, we
somtimes see the subsequent prctl which sets up the seccomp
filter and that happens to make the test work, but this is
unreliable and can lead to a weird situation where that prctl
got veto'd and thus rr's usual seccomp filter never got installed,
confusing rr down the line.

Fix all of that by switching the veto'd syscall to be the kill(SIGSTOP)
instead, which is present on all architectures.


Keno:
early_error passes on CI on this branch :tada:

Keno:
All aarch64 CI is green :)

yuyichao:
So it seems like the CI badge can be reenabled again.

Keno:
Yes, will do once this is merged.

PR #3317, opened 2022-07-04
Add initial Neoverse V1 and N2 support

yuyichao:
Tested on Amazon Graviton3 (V1). N2 support is based on known perf counter info and isn't tested on hardware.

SVE (registers and first/non fault load) not yet supported.


yuyichao:
All the tests passed but of course that's at least partially because we aren't testing for SVE and there's very limited (non-existing?) use of SVE in general userspace ATM.

PR #3318, opened 2022-07-04
Adding check and optimizations for thread local storage

yuyichao:
* Reduce copying of the thread local storage when possible
* Add static assertion to make sure the offset we hardcode in the assembly for `stub_scratch_2` remains valid.


yuyichao:
Passed ~30 test runs locally.

PR #3392, opened 2022-11-03
Make the artifact storage location configurable.

DanAlbert:
Forks shouldn't be attempting to push their artifacts to the upstream bucket. Make this location configurable in the cloud build config so each repository can have their own storage location.

This does also change the storage layout to avoid clobbering stored artifacts with each build. The output path as well as the bucket name could probably be made configurable instead if that behavior should be kept.

### DIFF
--- a/.android/cloudbuild.yaml
+++ b/.android/cloudbuild.yaml
@@ -5,5 +5,7 @@ steps:
 timeout: 3600s
 artifacts:
   objects:
-    location: 'gs://rr-android-testing_artifacts/rr'
+    # Configure _ARTIFACT_BUCKET in you cloud build trigger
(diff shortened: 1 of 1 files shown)


DanAlbert:
I'm not really sure how to test this before merging. I think you could probably go into your cloud build settings and add this substitution for PRs, and then re-run the checks? I copied this pattern from another project where I've used cloud builders: https://github.com/google/prefab/tree/master/.cloudbuild. (I don't store PR outputs there, but you can see it in post-submit and in release).

DanAlbert:
Failing tests appear to be unrelated.

khuey:
We're going to need to update the artifact path to contain the correct rr version (we're on 5.6 now)

khuey:
I think 0b5d4fedfe9ae0ad53b527adb041cc1e8dcf7f83 will do that

Issue #3394, opened 2022-11-09
cdparanoia: Unsupported ioctl: SG_EMULATED_HOST


chkno:
Hi.  https://rr-project.org/ says "Please [file github issues](https://github.com/rr-debugger/rr/issues/) for unsupported system calls."  This is that.

I get `Fatal register mismatch` while trying to replay a cdparanoia run.  The mismatch is in a `write` syscall in which cdparanoia is reporting the result of its device probing: In the original run it prints `Drive is ATAPI (using SG_IO host adaptor emulation)` but in the replay it tries to print `Drive is SCSI`. cdparanoia does this device detection and printing in [interface/scsi_interface.c](https://svn.xiph.org/trunk/cdparanoia/interface/scsi_interface.c):

```c
  cdmessage(d,"\nChecking for SCSI emulation...\n");

  if (ioctl(fd,SG_EMULATED_HOST,&atapiret)){
    cderror(d,"\tSG_EMULATED_HOST ioctl() failed!\n");
    return(-1);
  } else {
    if(atapiret==1){
      if(d->interface == SGIO_SCSI){
	cdmessage(d,"\tDrive is ATAPI (using SG_IO host adaptor emulation)\n");
      }else if(d->interface == SGIO_SCSI_BUGGY1){
	cdmessage(d,"\tDrive is ATAPI (using SG_IO host adaptor emulation with workarounds)\n");
      }else{
	cdmessage(d,"\tDrive is ATAPI (using SCSI host adaptor emulation)\n");
	/* Disable kernel SCSI command translation layer for access through sg */
	if (ioctl(fd,SG_SET_TRANSFORM,0))
	  cderror(d,"\tCouldn't disable kernel command translation layer\n");
      }
      d->is_atapi=1;
    }else{
      cdmessage(d,"\tDrive is SCSI\n");
      d->is_atapi=0;
    }
```

I get `Fatal register mismatch` trying to replay cdparanoia runs in rr versions 5.5.0, 5.6.0, and HEAD (currently 0b5d4fedfe9ae0ad53b527adb041cc1e8dcf7f83).

Log of attempt at HEAD (currently 0b5d4fedfe9ae0ad53b527adb041cc1e8dcf7f83):

```
[...]

PR #3395, opened 2022-11-09
Support some more SG_ ioctls

chkno:
Fixes https://github.com/rr-debugger/rr/issues/3394 .  Allows use with [cdparanoia](https://xiph.org/paranoia).


khuey:
It took me a minute to figure out that SG_SET_RESERVED_SIZE works because it (by happenstance) fits the regular ioctl pattern, but ok, this looks good.

PR #3403, opened 2022-11-29
Fix watchpoints on AArch64

khuey:
I don't think I can explain this any more concisely than I did in the code comment.


Keno (approved):
Seems like a strange quirk for the kernel to expose, but the implementation looks fine.

icecream95:
This seems to fix the issue for me, but I hit a build dependency error the first time I tried to compile it (`SyscallEnumsForTestsGeneric.generated` was not found)—I think `watchpoint_unaligned2` should depend on `Generated`?

khuey:
Yeah I think that's right. Done.

PR #3413, opened 2022-12-14
Adjust buildkite setup

staticfloat:
We've updated our buildkite tags a bit, this matches the new configuration

### DIFF
--- a/.buildkite/lib/generate.jl
+++ b/.buildkite/lib/generate.jl
@@ -42,7 +42,7 @@ function generate(platform::Platform)
                 "key" => job_key,
                 "timeout_in_minutes" => 45,
                 "agents" => Dict(
-                    "sandbox.jl" => "true",
+                    "sandbox_capable" => "true",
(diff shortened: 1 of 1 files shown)


staticfloat:
Please note that there was a hardware failure in the x86_64 runners we were using for x86_64 CI for this project.  We've swapped around to using AWS `c5d.metal` machines, but there appear to be some novel failures in the tests on those machines:

```
The following tests FAILED:
	460 - x86/pkeys (Failed)
	461 - x86/pkeys-no-syscallbuf (Failed)
	1886 - x86/pkeys-32 (Failed)
	1887 - x86/pkeys-32-no-syscallbuf (Failed)
```

khuey:
The pkeys failures are a known kernel bug.

DilumAluthge:
@Keno Should we mark the `pkeys` tests as "allowed to fail" in CI? Otherwise, every commit on master (and every PR) is going to have a red X, which can be annoying (and can distract from real test failures).

PR #3432, opened 2023-02-01
Fix rr build for newer NDKs.

DanAlbert:
The presence of the execinfo.h header does not mean that the backtrace APIs will be available at runtime on Android. The header is always available (to facilitate runtime detection of APIs via `dlsym`), but the API is only available for devices that are API 33 or newer.

This was not symptomatic with older NDKs because the execinfo.h header was not yet present (it was added in r24 or r25; rr was tested for Android with r23).

### DIFF
--- a/src/util.cc
+++ b/src/util.cc
@@ -33,6 +33,10 @@
 #include <random>
 #include <sstream>
 
+#ifdef __ANDROID__
+#include <android/api-level.h>
+#endif
+
 #include "preload/preload_interface.h"
 
 #include "AddressSpace.h"
@@ -49,6 +53,14 @@
 #include "kernel_metadata.h"
 #include "log.h"
 #include "seccomp-bpf.h"
+
+#ifdef __ANDROID__
+// backtrace() is not available in bionic until API 33. The presence of the
+// header does not imply the availability of APIs at runtime.
+#define CAN_USE_BACKTRACE_API (__ANDROID_MIN_SDK_VERSION__ >= 33)
+#else
+#define CAN_USE_BACKTRACE_API EXECINFO_H
+#endif
 
 void good_random(uint8_t* out, size_t out_len);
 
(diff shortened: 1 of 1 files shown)


khuey:
Hmm, if Android shipped an execinfo.h without backtrace(3) maybe we should just gate on having backtrace(3) instead of gating on having execinfo.h (and having some Android version specific logic here)?

If that sounds reasonable to you I'll make that change.

DanAlbert:
Yeah, if you prefer to write it that way, that works fine for me. I'll close this.

Android has `backtrace`, but not until `minSdkVersion >= 33` (`minSdkVersion` is the `-DANDROID_PLATFORM` CMake argument). We could alternatively raise that for rr since we're only worrying about the platform for the moment, and worry about NDK use cases later.

khuey:
Could you verify https://github.com/rr-debugger/rr/commit/b2104615d9c790a09e92332b119cfc9b9a831796 works for you?

DanAlbert:
Yes, that works for the configuration we currently use for Android builds. If we ever want to take advantage of https://github.com/android/ndk/issues/1003, it won't do the right thing. The callsite would have to change for that to work anyway though. We probably do want to do that eventually (to improve behavior on new devices even if the minimum supported OS is older), but that feature hasn't shipped yet so there's no rush.

khuey:
Do I understand correctly that android/ndk#1003 would require us to test that we can actually execute backtrace(3) rather than just find it in a header (because it might be a weak symbol that can't be resolved)?

We could switch to `check_c_source_runs` when that day comes.

DanAlbert:
> Do I understand correctly that https://github.com/android/ndk/issues/1003 would require us to test that we can actually execute backtrace(3) rather than just find it in a header (because it might be a weak symbol that can't be resolved)?

It will make all maybe-available-at-runtime symbols weak. That lets you link against them without causing the program to fail to load on older devices that don't have the API. At runtime you use `__builtin_available` to query the presence of the API. `-Werror=unguarded-availability` keeps you from accidentally calling these APIs without a check (which would segfault).

> We could switch to check_c_source_runs when that day comes.

That wouldn't help. For a case like this (`minSdkVersion 28`, API introduced in API 33, rr _might_ run on API 33, but also _might_ run on API 28), there is no way to know at build time whether the API will be available at run time. If you want to use new APIs that are available on only a subset of the OS versions your app supports, the only way to do that is with a runtime check.

But no, we don't need to worry about this today :) It's a pretty straightforward fix when the time comes.

PR #3455, opened 2023-03-02
Revert to minimum cmake version 3.1.

bernhardu:
And use add_compile_definitions just equal or above cmake 3.12.

Related: f39e6e9ea
Related: 797278698

Raising the minimum cmake version broke at least a vanilla Ubuntu 18.04
with 3.10.2, which worked before those commits, except the backtrace.
Maybe it is worth to leave the minimum version 3.1 for now?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 # *-* Mode: cmake; *-*
 
-cmake_minimum_required(VERSION 3.12.0)
+cmake_minimum_required(VERSION 3.1.0)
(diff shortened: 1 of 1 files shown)


khuey:
I think we should just use `add_definitions` instead of `add_compile_definitions`.

khuey:
Obsoleted by 6604017d8b04082e3a2b71570f165b3a35e8b717

bernhardu:
Thanks, that is way better.

PR #3460, opened 2023-03-11
[Do not merge] CI fd_cleanup test.

bernhardu:


bernhardu:
Seems this is no longer visible. Last appeareance was [1005](https://buildkite.com/julialang/rr/builds/1005#_). In the next 38 runs no fd_cleanup appeared.

PR #3469, opened 2023-03-27
CI: Allow manual retry of succesful builds

Keno:
By default, retry is only enabled for failed builds. When investigating intermittent failures, it can be useful to retry successful builds also, so enable that in the UI.

https://buildkite.com/docs/pipelines/command-step#retry-attributes-manual-retry-attributes

cc @staticfloat @DilumAluthge


DilumAluthge (approved):
Makes sense to me!

staticfloat (approved):

PR #3470, opened 2023-03-27
CI: Try relaxing architecture restriction on generator script

Keno:
The script that schedules the actual CI jobs is currently required to run on x86_64. When the x86_64 queues are backed up (like today), it would be nice to be able to run this on aarch64, to at least get some indication of whether the PR works. Try this and see if everything just magically works.


DilumAluthge (approved):

Keno:
Looks like this is working.

PR #3480, opened 2023-04-01
Avoid some warnings about possible uses of unintialized variables.

bernhardu:
Found with g++ 12.2.

### DIFF
--- a/src/Task.cc
+++ b/src/Task.cc
@@ -1191,7 +1191,9 @@ bool Task::set_aarch64_debug_regs(int, ARM64Arch::user_hwdebug_state *, size_t) 
   FATAL() << "Reached aarch64 code path on non-aarch64 system";
   return false;
 }
-bool Task::get_aarch64_debug_regs(int, ARM64Arch::user_hwdebug_state *) {
+bool Task::get_aarch64_debug_regs(int, ARM64Arch::user_hwdebug_state *regs) {
+  // Following memset just to silence a warning about dbg_info may be used uninitialized.
+  memset(regs, 0, sizeof(*regs));
   FATAL() << "Reached aarch64 code path on non-aarch64 system";
   return false;
 }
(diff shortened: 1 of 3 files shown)


src/Task.cc
```diff
@@ -1191,7 +1191,9 @@ bool Task::set_aarch64_debug_regs(int, ARM64Arch::user_hwdebug_state *, size_t)
   FATAL() << "Reached aarch64 code path on non-aarch64 system";
   return false;
 }
-bool Task::get_aarch64_debug_regs(int, ARM64Arch::user_hwdebug_state *) {
+bool Task::get_aarch64_debug_regs(int, ARM64Arch::user_hwdebug_state *regs) {
```
khuey, line 1194: This seems like a ridiculous warning to me.

bernhardu: The warning manifests at the call of the function. See the full warning text below.
I did not want to really initialize the variable at declaration, to avoid issues in code I cannot even test.
As this function should never be called in x86_64, I guessed this change in "dead" code
should be at least not harmful.
```c++
In function ‘void rr::query_max_bp_wp(Task*, ssize_t*, ssize_t*)’,
    inlined from ‘bool rr::set_debug_regs_arch(Task*, const Task::HardwareWatchpoints&) [with Arch = ARM64Arch]’ at .../rr/src/Task.cc:1738:20:
.../rr/src/Task.cc:1724:17: warning: ‘bps.rr::ARM64Arch::user_hwdebug_state::dbg_info’ may be used uninitialized [-Wmaybe-uninitialized]
 1724 |   *max_bp = bps.dbg_info & 0xff;
      |             ~~~~^~~~~~~~
.../rr/src/Task.cc: In function ‘bool rr::set_debug_regs_arch(Task*, const Task::HardwareWatchpoints&) [with Arch = ARM64Arch]’:
.../rr/src/Task.cc:1719:33: note: ‘bps’ declared here
 1719 |   ARM64Arch::user_hwdebug_state bps;
      |                                 ^~~
```

PR #3491, opened 2023-04-11
ASan: Run with verify_asan_link_order=0.

bernhardu:
Avoids error message:
```
  ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
```

Affected tests:
```
  nested_detach_kill
  nested_detach_kill_stuck
  nested_detach_wait
```
(These tests may also need the `ASan: Copy just used pages from ANONYMOUS|NORESERVE mappings.` patch too.)

### DIFF
--- a/src/RecordSession.cc
+++ b/src/RecordSession.cc
@@ -2223,6 +2223,8 @@ static ExeInfo read_exe_info(const string& exe_file) {
   ElfFileReader reader(fd);
   ret.arch = reader.arch();
 
+  // Don't try to modifiy LD_PRELOAD when rr itself is built with ASan
+#ifndef __SANITIZE_ADDRESS__
   DynamicSection dynamic = reader.read_dynamic();
   for (auto& entry : dynamic.entries) {
     if (entry.tag == DT_NEEDED && entry.val < dynamic.strtab.size()) {
@@ -2236,6 +2238,7 @@ static ExeInfo read_exe_info(const string& exe_file) {
       }
     }
   }
+#endif
 
   auto syms = reader.read_symbols(".dynsym", ".dynstr");
   for (size_t i = 0; i < syms.size(); ++i) {
--- a/src/util.cc
+++ b/src/util.cc
@@ -2461,6 +2461,10 @@ int __lsan_is_turned_off(void)
 {
   return rr_lsan_is_turned_off;
 }
+#include <sanitizer/asan_interface.h>
+const char *__asan_default_options() {
+  return "verify_asan_link_order=0";
+}
 #endif
 
 bool coredumping_signal_takes_down_entire_vm() {


Keno:
I feel like this conflates the issues of rr itself running under asan and rr tracing something running under asan (which itself may be rr). If it's just a matter of fixing the tests, I think this should be in `__asan_default_options` so it doesn't propagate to children.

bernhardu:
Thanks, let me try how it works when I put it into the default options.

bernhardu:
For the tests `nested_detach_kill`, `nested_detach_kill_stuck` and `nested_detach_wait` following diff would be sufficient.

```diff
diff --git a/src/main.cc b/src/main.cc
index 128fcbe3..f7ea5c34 100644
--- a/src/main.cc
+++ b/src/main.cc
@@ -231,6 +231,13 @@ size_t saved_argv0_space() {
 
 } // namespace rr
 
+#ifdef __SANITIZE_ADDRESS__
+#include <sanitizer/asan_interface.h>
+const char *__asan_default_options() {
+  return "verify_asan_link_order=0";
+}
+#endif
+
 using namespace rr;
 
 int main(int argc, char* argv[]) {
```

But the test `nested_detach` still fails.
Following is the content of an example recording:
```
$ bin/rr record bin/rr record --nested=detach bin/simple
rr: Saving execution to trace directory `/home/bernhard/.local/share/rr/rr-23'.
rr: Saving execution to trace directory `/home/bernhard/.local/share/rr/simple-34'.
==1247576==ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.

$ bin/rr ps rr-23
PID     PPID    EXIT    CMD
1247550 --      detach  bin/rr record --nested=detach bin/simple

$ bin/rr replay --debugger-option=-q rr-23
(rr) b main
(rr) cont
(rr) print *environ@100
... "LD_PRELOAD=libasan.so.8:/home/bernhard/x86_64_asan/obj/bin/../lib/rr/librrpreload.so::::libdl.so.2", ...

$ bin/rr ps simple-34
PID     PPID    EXIT    CMD
1247576 --      1       bin/simple

$ bin/rr replay --debugger-option=-q simple-34
(rr) cont
(rr) print *environ@100
... "LD_PRELOAD=/home/bernhard/x86_64_asan/obj/bin/../lib/rr/librrpreload.so::::libdl.so.2:libasan.so.8:libdl.so.2", ...
```

So I guess the question is here if the simple binary should have the "libasan.so.8" in LD_PRELOAD?

To make it even more complicated the `nested_detach` 32-bit is also subject of another patch bf7a42acea which I used for above test.

bernhardu:
I pushed a new version with no longer setting the environment, instead using `__asan_default_options`.

bernhardu:
Changes:
- Rebased to current tip.
- Do not try to add libasan to LD_PRELOAD below ASan-rr.

bernhardu:
Just a rebase to current tip.

PR #3501, opened 2023-04-15
[Do not merge] sioc still fails at exclusive-amdci1.0

bernhardu:


bernhardu:
Hello @Keno, hello @DilumAluthge, is there a way to run CI on a specific CI node/agent?
I would like to run it specifically at exclusive-amdci1.0.
There a few errors show up that seem specific to this node, e.g. https://buildkite.com/julialang/rr/builds/1087.

Keno:
You can set the `"exclusive" => "true"` tag here: https://github.com/rr-debugger/rr/blob/master/.buildkite/lib/generate.jl#L45 to force an exclusive agent.

bernhardu:
Thanks, I tried it but I am not sure where to feed in the agent name. I tried it with a "name" element below, but the the job does not start. But the desired agent might just be busy and I am to unpatient?

Keno:
You can't target a specific agent by name. If it doesn't show up on the other exclusive agents, I think it can be forced manually, but there's not too many of those agents anyway, so you should get it after a few tries

DilumAluthge:
The exclusive agents often have a very long backlog.

bernhardu:
Hello, thanks for the input. Pushing today with just the `"exclusive" => "true"` got me almost every time at amdci1. I hope I did not generate too much noise.

I guess following might be what I was looking for, but because the environment entry `BUILDKITE_AGENT_META_DATA_HOSTNAME` does not appear it would need another [tag on the agent configuration](https://buildkite.com/docs/agent/v3/cli-start#run-a-job-on-the-agent-that-uploaded-it-also-known-as-node-affinity) to make something called "node-affinity" possible. If the agents are configured that way, wouldn't then a `"hostname" => "amdci1"` do it?

bernhardu:
Got handled in #3503, closing this test PR.

PR #3513, opened 2023-05-03
support for Intel SapphireRapid

byrnedj:
This adds support for Intel Xeon 4th gen codename SapphireRapids. Let me know if there are any issues.

### DIFF
--- a/src/PerfCounters.cc
+++ b/src/PerfCounters.cc
@@ -86,7 +86,8 @@ enum CpuMicroarch {
   IntelRocketlake,
   IntelAlderlake,
   IntelRaptorlake,
-  LastIntel = IntelRaptorlake,
+  IntelSapphireRapid,
+  LastIntel = IntelSapphireRapid,
   FirstAMD,
   AMDF15R30 = FirstAMD,
   AMDZen,
@@ -155,6 +156,7 @@ struct PmuConfig {
 // See Intel 64 and IA32 Architectures Performance Monitoring Events.
 // See check_events from libpfm4.
 static const PmuConfig pmu_configs[] = {
+  { IntelSapphireRapid, "Intel SapphireRapid", 0x5111c4, 0, 0, 125, PMU_TICKS_RCB },
   { IntelRaptorlake, "Intel Raptorlake", 0x5111c4, 0, 0, 125, PMU_TICKS_RCB },
   { IntelAlderlake, "Intel Alderlake", 0x5111c4, 0, 0, 125, PMU_TICKS_RCB },
   { IntelRocketlake, "Intel Rocketlake", 0x5111c4, 0, 0, 100, PMU_TICKS_RCB },
--- a/src/PerfCounters_x86.h
+++ b/src/PerfCounters_x86.h
@@ -85,6 +85,8 @@ static CpuMicroarch compute_cpu_microarch() {
       return IntelAlderlake;
     case 0xb0670:
       return IntelRaptorlake;
+    case 0x806f0:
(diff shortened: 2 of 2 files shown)


khuey:
I assume you ran `make check` and everything passes?

byrnedj:
Hi, actually it seems I need to add support for Intel CET for the SapphireRapid platform.

I added the following patch - but it appears to be more complicated than just letting the sys 
call go through. Would you have any suggestions on how to support CET from `arch_prctl` calls?

```
diff --git a/src/kernel_supplement.h b/src/kernel_supplement.h
index 204d57d7..175f5da6 100644
--- a/src/kernel_supplement.h
+++ b/src/kernel_supplement.h
@@ -387,6 +387,24 @@ struct rr_input_mask {
 #define ARCH_SET_CPUID 0x1012
 #endif
 
+// support for control flow enforcement
+// on Intel processors
+#ifndef ARCH_CET_STATUS
+#define ARCH_CET_STATUS 0x3001
+#endif
+#ifndef ARCH_CET_DISABLE
+#define ARCH_CET_DISABLE 0x3002
+#endif
+#ifndef ARCH_CET_LOCK
+#define ARCH_CET_LOCK 0x3003
+#endif
+#ifndef ARCH_CET_ALLOC_SHSTK
+#define ARCH_CET_ALLOC_SHSTK 0x3004
+#endif
+#ifndef ARCH_CET_LEGACY_BITMAP
+#define ARCH_CET_LEGACY_BITMAP 0x3005
+#endif
+
 // New in the 4.15 kernel
 #ifndef MAP_SYNC
 #define MAP_SYNC  0x80000
diff --git a/src/record_syscall.cc b/src/record_syscall.cc
index 53d9263f..8dbd348a 100644
--- a/src/record_syscall.cc
+++ b/src/record_syscall.cc
@@ -4754,6 +4754,14 @@ static Switchable rec_prepare_syscall_arch(RecordTask* t,
           break;
         }
 
+        case ARCH_CET_STATUS:
+        case ARCH_CET_DISABLE:
+        case ARCH_CET_LOCK:
+        case ARCH_CET_ALLOC_SHSTK:
+        case ARCH_CET_LEGACY_BITMAP: {
+          break;
+        }
+
         default:
           syscall_state.expect_errno = EINVAL;
           break;
```

rocallahan:
CET is likely to be a lot of work to support.

is it possible to turn off CET and test rr?

khuey:
Maybe I'm missing something but I can't find CET support in the mainline Linux kernel ...?

byrnedj:
> Maybe I'm missing something but I can't find CET support in the mainline Linux kernel ...?

Yes, I realized that I was on a development kernel. I have reverted to a kernel without CET.

With that said, I do get a few failures. I'll try an upstream kernel just to be sure.

Screenshot 2023-05-08 at 3 24 39 PM

Digging into test 471, here is the rr replay - do you have any thoughts on what might be causing the assert to fail.

Screenshot 2023-05-08 at 3 24 21 PM

khuey:
What kernel version did you revert to? There's a known kernel bug that affected the pkeys tests.

byrnedj:
Hi,

Sorry for the delay, I upgraded to 6.2.12-1.el8.elrepo.x86_64 (from CentOS repo). The only two tests that fail now are:

image

Are these ones the known issues?

khuey:
I'm not sure what's going on with those two tests but as you note they're failing on CI, and those tests are definitely not hardware specific. I think everything is fine with Sapphire Rapids.

PR #3540, opened 2023-06-22
Add support for PTRACE_O_TRACEVFORKDONE

### LINKED ISSUE #3524: Support PTRACE_O_TRACEVFORKDONE
lldb wants this.

theIDinside:
This PR closes #3524  

This fix seemed at first too easy to be true and though I might have missed something, after discussion with @khuey it seemed to us that this solution will work.

TRACEVFORK and TRACEVFORKDONE events are delivered to the parent process (the vfork parent). As such, this PR queues up (emulated) a TRACEVFORKDONE in `rec_process_syscall_arch` for `Arch::vfork`, because according to VFORK specs, the parent will be suspended for the duration until a TRACEVFORKDONE event is delivered, either because the vforked child `exit`s or `exec`s.

We also talked about having the emulated ptrace stop happen on exits and execs for a child, but that would involve the child having to know if it was a `vfork child` and track who is its parent so that it can emulate a ptrace stop. We landed in the conclusion for now that this was probably a worse idea and potentially more fragile (more things can go wrong). But I figured I'd write about this discussion for posterity.


src/record_syscall.cc
```diff
@@ -6393,8 +6393,20 @@ static void rec_process_syscall_arch(RecordTask* t,
   // syscall completes --- and that our TaskSyscallState infrastructure can't
   // handle.
   switch (syscallno) {
+    case Arch::vfork: {
+      if (t->emulated_ptrace_options & PTRACE_O_TRACEVFORKDONE) {
+        t->emulate_ptrace_stop(
+            WaitStatus::for_ptrace_event(PTRACE_EVENT_VFORK_DONE));
+      }
+      // see Arch::clone, Arch::vfork used to fall-through there.
```
khuey, line 6401: Why can't it continue to fall through here?

src/test/vfork_setopts.c
```diff
@@ -0,0 +1,35 @@
...
+  switch (fork_child) {
+    case 0: {
+      assert(1 == read(EXECUTION_FENCE_PIPES[0], &notify, 1) &&
+             "Failed to read notify event");
+      exit(66);
+    } break;
+    default:
+      attach(fork_child);
+      test_assert(1 == write(EXECUTION_FENCE_PIPES[1], "1", 1));
+      // at this point, we can now set otps, because ATTACH ptrace stop event
```
khuey, line 24: nit: opts

src/record_syscall.cc
theIDinside: I think you told me to separate it (or maybe I had misunderstood)? I had initially the `[[fallthrough]]` attribute there, so I had thought maybe that was not according to convention.

I think the `[[fallthrough]]` attribute is ok, in c++14, right? That way we won't get any warnings about it.

khuey: Yeah I don't think I understood how simple this was when I told you that :)

We use RR_FALLTHROUGH IIRC.

theIDinside: Great :)

theIDinside:
@khuey I've also strengthened the test in the last commit. I wanted to be sure that, let's say someone sends the vfork parent a signal during it's suspension (waiting for vfork child to `exit` or `exec`) - that it comes *after* the `VFORK_DONE`. So this test verifies that. So I think this solution actually works safely, unless there's some really, really strange edge case.

khuey (approved):

PR #3554, opened 2023-07-07
[Do not merge] CI test for madvise_dontfork.

bernhardu:
Issue 3545.


.buildkite/lib/generate.jl
```diff
@@ -7,6 +7,13 @@ function generate(platform::Platform)
     echo "--- Print kernel information"
     uname -a
 
+    echo "--- Print environment variables"
```
GitMensch, line 10: Would it be useful to include this in general? Then a separate PR would be reasonable.

PR #3558, opened 2023-07-16
Add rr commands to remove and move traces

vinhill:
As discussed in #1415, this pull request implements `rr rm` and `rr mv` commands to manage traces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -555,6 +555,7 @@ set(RR_SOURCES
   src/MmappedFileMonitor.cc
   src/MonitoredSharedMemory.cc
   src/Monkeypatcher.cc
+  src/MvCommand.cc
   src/PackCommand.cc
   src/PerfCounters.cc
   src/PidFdMonitor.cc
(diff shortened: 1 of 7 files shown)


vinhill:
It seems for c++14, libstdc++ and libc++ behave differently around the filesystem library.
- clang does not work with the gcc-11 version of the filesystem implementation in libstdc++, see [bug](https://bugs.launchpad.net/ubuntu/+source/gcc-11/+bug/1995569)
- libc++ deprecates `std::experimental::filesystem` and refers to `std::filesystem`, though I can only find `std::__fs::filesystem`.

I'm unsure what to do about this. For the first problem, we can try to disable `Wmissing-exception-spec` or have clang use libc++. For the second one, either a macro for selecting the namespace based on the compiler or defining `_LIBCPP_NO_EXPERIMENTAL_DEPRECATION_WARNING_FILESYSTEM`. I can also try to work without the filesystem library.

vinhill:
I think I prefer not using filesystem, as I cannot get clang to work with libstdc++ nor to link correctly against the filesystem library.

PR #3576, opened 2023-08-15
Add `--save-as` param to record

theIDinside:
`--save-as` specifies the trace dir name (basename?), not the full path.

Merged `output_trace_dir` in RecordFlags with `name` into new type `TraceOutputPath`.

Changed all usage of `output_trace_dir` string to use this type instead.

This PR is backwards compatible, in the sense that the old usage of the `-o` flag remains the same, if `--save-as` is not provided, i.e., will error out, if dir exists etc. If `--save-as` is provided together with `-o` the new behavior will happen instead, where the output dir will be the "root dir", thus, the user can save many traces there.

If only `--save-as` is provided, record uses normal behavior, of setting the trace root dir to $RR_TRACE_DIR (or it's user provided env var).

Naming of user provided dirs, follows old behavior, i.e. appending -0, -1, -2 etc to the trace dir. I think this is preferable - if some automated thing is recording something with a specific name provided, this makes it so the end user don't have to manage the file system (i.e. checking if that name is "taken" and having to do clean up before recording, etc.) - especially since there's kind of functionality for that, already existing (just using the `-o` on it's own, though you have to specify a path, not a name)

### DIFF
--- a/src/RecordCommand.cc
+++ b/src/RecordCommand.cc
@@ -61,6 +61,9 @@ RecordCommand RecordCommand::singleton(
     "                             _RR_TRACE_DIR gets ignored.\n"
     "                             Directory name is given name, not the\n"
     "                             application name.\n"
+    "  --save-as=<NAME>           Name of the new recording's directory. If\n"
+    "                             a recording with that name already exists, normal\n"
+    "                             number appending is applied."
     "  -p --print-trace-dir=<NUM> print trace directory followed by a newline\n"
     "                             to given file descriptor\n"
     "  --syscall-buffer-sig=<NUM> the signal used for communication with the\n"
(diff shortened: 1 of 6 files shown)


theIDinside:
As a side note @khuey & @rocallahan - I noticed that in the `RecordFlags` `-s` is *sort of* taken - though it's not parsed by the option parsing logic. It's used for `always_switch` - but if it's not parsed in the option parsing logic, how can it ever be set? 

I had originally intended the `--save-as` to have `-s` as a short, when I found that it was being used (but not being parsed). Perhaps this needs fixing?

khuey:
> though it's not parsed by the option parsing logic

Why do you think it's not parsed?

GitMensch:
@theIDinside Would you mind rebasing this PR so it is more easily tested and integrated?

theIDinside:
> @theIDinside Would you mind rebasing this PR so it is more easily tested and integrated?

Absolutely. I need to do the same with another PR so I will do that after work or this weekend.

theIDinside:
> @theIDinside Would you mind rebasing this PR so it is more easily tested and integrated?

Rebased & added tests for good measure.

theIDinside:
I screwed this one up. I will re-send PR.

PR #3405, opened 2022-11-30
Add Intel Tremont

Frederick888:
The cpu_type is from an Intel N6005 [1].

[1] https://www.intel.com/content/www/us/en/products/sku/212327/intel-pentium-silver-n6005-processor-4m-cache-up-to-3-30-ghz/specifications.html

---

Recently got my hands on a new board with this chip. Since it's a successor
of Goldmont Plus, I gave the same configuration a shot and it looked ok(?)

Failed tests:

```
The following tests FAILED:
	468 - prctl (Failed)
	469 - prctl-no-syscallbuf (Failed)
	724 - sioc (Failed)
	725 - sioc-no-syscallbuf (Failed)
	1140 - x86/string_instructions_replay (Failed)
	1141 - x86/string_instructions_replay-no-syscallbuf (Failed)
	1144 - x86/string_instructions_watch (Failed)
	1145 - x86/string_instructions_watch-no-syscallbuf (Failed)
	1388 - x86/string_instructions_break (Failed)
	1389 - x86/string_instructions_break-no-syscallbuf (Failed)
	1390 - x86/string_instructions_replay_quirk (Failed)
	1391 - x86/string_instructions_replay_quirk-no-syscallbuf (Failed)
	1892 - prctl-32 (Failed)
	1893 - prctl-32-no-syscallbuf (Failed)
	2564 - x86/string_instructions_replay-32 (Failed)
	2565 - x86/string_instructions_replay-32-no-syscallbuf (Failed)
	2568 - x86/string_instructions_watch-32 (Failed)
	2569 - x86/string_instructions_watch-32-no-syscallbuf (Failed)
	2812 - x86/string_instructions_break-32 (Failed)
	2813 - x86/string_instructions_break-32-no-syscallbuf (Failed)
	2814 - x86/string_instructions_replay_quirk-32 (Failed)
	2815 - x86/string_instructions_replay_quirk-32-no-syscallbuf (Failed)

```

Full log: [rr_check.log](https://github.com/rr-debugger/rr/files/10121994/rr_check.log)

PR #3624, opened 2023-10-01
Add support for Celeron N5095

ttsiodras:
One line patch I needed to do to support my N5095.


GitMensch:
Just for reference: The _older_ PR that would have brought this in (completely with the missing arch) is #3405.
@ttsiodras because in the referenced PR a bunch of tests (especially string_instructions) do fail: What are the results of [running the testsuite](https://github.com/rr-debugger/rr/wiki/Building-And-Installing#tests) on your machine?

PR #3636, opened 2023-10-25
Be more flexible selecting an UTF-8 locale.

bernhardu:
The current used en_US.UTF-8 is not guaranteed to be available in e.g. Debian, which has for this reason C-UTF-8.
The Debian downstream maintainer has for this a patch in the [rr sources](https://sources.debian.org/src/rr/5.7.0-2/debian/patches/test-locale.patch/).
But this C.UTF-8 seems not to be guaranteed to be available on systems with GNU libc < 2.35, [link](https://unix.stackexchange.com/questions/597962/how-widespread-is-the-c-utf-8-locale).

Therefore this attempt to select an existing UTF-8 locale.
This changes also the default from `LC_ALL=en_US.UTF-8` to `LC_ALL=C`.

See discussion at the end of https://github.com/rr-debugger/rr/issues/3545.

CC: @GitMensch

### DIFF
--- a/src/test/unicode.run
+++ b/src/test/unicode.run
@@ -1,2 +1,3 @@
 source `dirname $0`/util.sh
+set_utf_locale
 debug_test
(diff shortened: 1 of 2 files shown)


GitMensch (approved):
LGTM - thanks for this clean change!

PR #3639, opened 2023-10-31
[Do not merge] nproc test

bernhardu:
I wondered why vsyscall_reverse_next-32 takes in CI kind of long.
And suspect this has something to do with JULIA_CPU_THREADS=128
at a CPU supporting 64 threads.


bernhardu:
Because rr uses now the Github Actions CI instead of buildkite (and the lack of progress/time)
this PR is useless now. Closing.

PR #3663, opened 2023-12-05
bpf_query: fix build with older kernels

multun:


khuey:
That works for me, thanks.

PR #3692, opened 2024-02-18
CI: use a more specific name for the "Build Android" job

DilumAluthge:
Currently this job shows up as "Build", which I think is a little confusing. So this PR renames that job to "Build Android", which is what that job is specifically testing.


DilumAluthge:
cc: @Keno

PR #3775, opened 2024-07-02
pack test: be explicit about using /bin/chmod

sidkshatriya:
pack test: be explicit about using `/bin/chmod` as there is also a test binary named `chmod`


sidkshatriya:
(detected issue while figuring out #3774)

PR #3776, opened 2024-07-03
AVX512 Support

### LINKED ISSUE #2974: gdb integration doesn't support AVX-512
Something I noticed while mucking around with the PKRU stuff yesterday is that none of our gdb integration supports the AVX 512 registers (the ZMMs and XMM16+). I don't plan to implement this but I want it on file.

theIDinside:
Fixes #2974 

This PR introduces AVX512 support in replays.

CMake Changes:
Added the AVX512F_SUPPORT flag to CMake. Set by checking
if system supports AVX512F. If it doesn't it will not build (and therefore run) any of the AVX512 tests

Added the target descriptions. Assumed that PKRU support also implies AVX512 support as I could find no information that said otherwise.

Added gdb_avx512 test - sets 3 ZMM registers and reads them back in a gdb session of a replay to verify the contents.

We place mask registers K0-K7 after ZMM31H internally, because this simplifies the logic that parses the XSAVE area. The target description we provide, takes this into account (and also places them last) so they come after the ZMM registers in the 'g' packet. GDB doesn't care so long as we do what we say in the target desc

Changed test harness util.sh to also export TESTNAME variable as we may (this one test do at least) want to query from the python script what mode we are running in (64-bit/32-bit). See gdb_avx512.py for further info.

Not tested or developed for:
- Being able to write to / alter AVX512 registers during replay/diversion.


khuey:
I'll look at this in detail this week.

The assumption that AVX-512 implies PKRU is not necessarily correct because these two features can be controlled independently by the kernel. So it's possible to have a system with AVX-512 available but PKRU unavailable even if the silicon supports both. I do wonder if we could start pretending that certain things are available when they're not (e.g. for reads returning default register values and for writes returning errors) to avoid a combinatorial explosion in the number of target description XML files we need to ship. We're already up to 20 files. Any thoughts @rocallahan?

theIDinside:
> I'll look at this in detail this week.
> 
> The assumption that AVX-512 implies PKRU is not necessarily correct because these two features can be controlled independently by the kernel. So it's possible to have a system with AVX-512 available but PKRU unavailable even if the silicon supports both. I do wonder if we could start pretending that certain things are available when they're not (e.g. for reads returning default register values and for writes returning errors) to avoid a combinatorial explosion in the number of target description XML files we need to ship. We're already up to 20 files. Any thoughts @rocallahan?

GDB has a way to signal "could not read", which means, we could potentially do something like this:

Have a "one description to rule them all" (ish) kind of thing, and then just have RR splat 'x' in all the right places. Sure, it will mean that when reporting registers an architecture doesn't even have, we are doing the extra work of filling (N*regWidth + ...) bytes of 'x' into the g-packet, so that has to be taken into consideration.

At least this could be done for the PKRU (and potentially similar CPU features) feature, as it doesn't involve all that much unnecessary data.

theIDinside:
I guess we can assume the presence of AVX & AVX2 too, since these are very common features these days anyhow.

rocallahan:
> We're already up to 20 files.

We could dynamically assemble the top-level XML file. I.e. given the base CPU type (aarch64, i386, amd64) and some set of optional features (a subset of `avx`, `pkeys`, soon `avx512`), we report a register description file name of `<cpu>-<feature1>-...<featureN>-linux.xml` but don't actually ship all those files with rr. Instead when the client requests  file of that form, we can send dynamically generated contents with the `xi:include`s corresponding to the feature tags in the file name.

rocallahan:
> > We're already up to 20 files.
> 
> We could dynamically assemble the top-level XML file. I.e. given the base CPU type (aarch64, i386, amd64) and some set of optional features (a subset of `avx`, `pkeys`, soon `avx512`), we report a register description file name of `<cpu>-<feature1>-...<featureN>-linux.xml` but don't actually ship all those files with rr. Instead when the client requests file of that form, we can send dynamically generated contents with the `xi:include`s corresponding to the feature tags in the file name.

theIDinside:
> > We're already up to 20 files.
> 
> We could dynamically assemble the top-level XML file. I.e. given the base CPU type (aarch64, i386, amd64) and some set of optional features (a subset of `avx`, `pkeys`, soon `avx512`), we report a register description file name of `<cpu>-<feature1>-...<featureN>-linux.xml` but don't actually ship all those files with rr. Instead when the client requests a file of that form, we can send dynamically generated contents with the `xi:include`s corresponding to the feature tags in the file name.

Great idea. I think GDB does something similar under the hood. I can add this feature to this PR, or in a later PR, if we're all in agreement. What say you @khuey || @rocallahan ?

The feature would basically involve:

- Do the feature checking we already do
- Change `target_description_name` to `target_feature_set` to instead return a `Set<CPUFeature /* the GdbServerConnection::enum */>` as these are what determines what features are represented in the final `target.xml`
- Using the returned set, hand-serialize a `target.xml` with all the `<xi:include href="feature.xml">` 
- send over the wire

rocallahan:
> I think GDB does something similar under the hood. I can add this feature to this PR, or in a later PR, if we're all in agreement.

I think we should do it before this PR.

theIDinside:
> > I think GDB does something similar under the hood. I can add this feature to this PR, or in a later PR, if we're all in agreement.
> 
> I think we should do it before this PR.

Ok, can I wrap it up in the same PR or create a separate PR for it?

rocallahan:
Separate PR I think, thanks

src/GdbServerConnection.cc
```diff
@@ -87,10 +87,14 @@ static uint32_t get_cpu_features(SupportedArch arch) {
       auto cpuid_data = cpuid(CPUID_GETEXTENDEDFEATURES, 0);
       if ((cpuid_data.ecx & PKU_FEATURE_FLAG) == PKU_FEATURE_FLAG) {
         // PKU (Skylake) implies AVX (Sandy Bridge).
-        cpu_features |= GdbServerConnection::CPU_AVX | GdbServerConnection::CPU_PKU;
+        cpu_features |= GdbServerConnection::CPU_AVX | GdbServerConnection::CPU_AVX512 | GdbServerConnection::CPU_PKU;
```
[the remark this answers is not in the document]
khuey: No, because these can be independently disabled (like we tell Pernosco customers to do, for instance).

CMakeLists.txt
```diff
@@ -101,6 +101,8 @@ if (SUPPORTS_MACRO_PREFIX_MAP)
   set(FLAGS_COMMON "${FLAGS_COMMON} -fmacro-prefix-map=${CMAKE_SOURCE_DIR}/=")
 endif()
 
+try_compile(AVX512F_SUPPORT ${CMAKE_SOURCE_DIR} ${CMAKE_SOURCE_DIR}/src/test/gdb_avx512.c CMAKE_FLAGS -DCOMPILE_DEFINITIONS=-march=native)
```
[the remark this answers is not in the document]
theIDinside: I currently do not have (non-trivial) access to a machine without avx512f, but I figured using try_compile, we won't get build errors on machines without avx512f support. Maybe this is not needed? (`AVX512F_SUPPORT` is later checked, when adding the test, so if the platform can't succesfully compile it, it's not going to be tested either).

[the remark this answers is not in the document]
khuey: If you have a Linux machine you can get a machine without AVX-512 by putting clearcpuid=304 in your kernel boot parameters.

src/GdbServerConnection.cc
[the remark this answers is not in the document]
theIDinside: > In general nothing implies anything because these feature sets can be independently disabled. We tell Pernosco customers to disable AVX-512 all the time ...

I'll fix this assumption (I think you mentiond this months ago as well).

src/ExtraRegisters.cc
```diff
@@ -66,17 +66,59 @@ static bool reg_in_range(GdbServerRegister regno, GdbServerRegister low, GdbServ
...
+
+// Every range of registers (except K0-7) are 16 registers long. We use this fact to build
+// a lookup table, for the AVX2 and AVX512 registers.
+static bool reg_is_avx2_or_512(GdbServerRegister reg, RegData& out) noexcept {
+  if(reg < DREG_64_YMM0H || reg > DREG_64_K7) {
+    return false;
+  }
+
+  const auto selector = (reg - DREG_64_YMM0H) >> 4;
+  DEBUG_ASSERT(selector >= 0 && selector <= 5 && "GdbServerRegister enum values has been changed.");
```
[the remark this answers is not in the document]
theIDinside: Sure, I can change it to something like

```cpp
static bool reg_is_avx2_or_512(GdbServerRegister reg, RegData& out) noexcept {
  if (reg < DREG_64_YMM0H || reg > DREG_64_K7) {
    return false;
  }

  uint64_t index = UINT64_MAX;
  if (reg >= DREG_64_YMM0H && reg < DREG_64_XMM16) {
    index = 0;
  } else if ( reg>= DREG_64_XMM16 && reg <DREG_64_YMM16H) {
    index = 1;
  } else if (reg >= DREG_64_YMM16H && reg < DREG_64_ZMM0H) {
    index = 2;
  } else if (reg >= DREG_64_ZMM0H && reg < DREG_64_ZMM16H) {
    index = 3;
  } else if (reg >= DREG_64_ZMM16H && reg < DREG_64_K0) {
    index = 4;
  } else if (reg >= DREG_64_K0 && reg < (DREG_64_K7+1)) {
    index = 5;
  }

  if (index == UINT64_MAX) {
    return false;
  }

  const auto cfg = register_config_lookup_table[index];
  out.xsave_feature_bit = cfg.feature;
  out.size = cfg.size;

  // only YMM16-31 and ZMM16-31 have a base offset (16 and 32 respectively)
  const auto base_offset = ((index == YMM16_31) || (index == ZMM16_31)) ? cfg.size : 0;
  out.offset = cfg.register_offset(reg, base_offset);
  return true;
}
```

I think I did the selector approach to cut down on the branching, this is probably easier to follow though.

CMakeLists.txt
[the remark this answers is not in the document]
theIDinside: Thanks! I'll do that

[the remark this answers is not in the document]
rocallahan: Aren't we mixing up compiler support with AVX512 support here?
Whether gcc/clang can compile the test should depend on compiler version, not whether your CPU supports AVX512. I think you should be able to build the test successfully even if your CPU doesn't have AVX512.

src/ExtraRegisters.cc
```diff
@@ -66,17 +66,92 @@ static bool reg_in_range(GdbServerRegister regno, GdbServerRegister low, GdbServ
...
+    index = 2;
+  } else if (reg >= DREG_64_ZMM0H && reg < DREG_64_ZMM16H) {
+    index = 3;
+  } else if (reg >= DREG_64_ZMM16H && reg < DREG_64_K0) {
+    index = 4;
+  } else if (reg >= DREG_64_K0 && reg < (DREG_64_K7 + 1)) {
+    index = 5;
+  }
+
+  const auto cfg = register_config_lookup_table[index];
```
rocallahan, line 145: It would be a little cleaner if `RegisterConfig` stored the upper bound as well as the base; then you could just search the table to find the right entry instead of computing the index with that big if chain.

CMakeLists.txt
```diff
@@ -2014,6 +2024,11 @@ if(BUILD_TESTS)
                                   PROPERTIES COMPILE_FLAGS "-m32 ${RR_TEST_FLAGS}")
     endforeach(test)
 
+    if (x86ish)
+      configure_file("${CMAKE_CURRENT_SOURCE_DIR}/src/test/x86/gdb_avx512.c" "${CMAKE_CURRENT_BINARY_DIR}/32/gdb_avx512.c" COPYONLY)
+      set_source_files_properties("${CMAKE_CURRENT_BINARY_DIR}/32/${test}.c" PROPERTIES COMPILE_FLAGS "-m32 ${RR_TEST_FLAGS} -g3 -mavx512f")
```
rocallahan, line 2029: Do you really need `-mavx512f`? What happens if you remove that (and make this a regular test), and remove the `defined(__AVX512F__)` code in the test? I think it will probably just work? That test only uses AVX512 via inline assembly which I don't think requires `-mavx512f`.

rocallahan (changes requested):

src/ExtraRegisters.cc
theIDinside: right, I get what you mean, I'll do that!

theIDinside: I'm also going to rename it to `RegisterDescriptor` as I believe this better describes what it's doing.

CMakeLists.txt
theIDinside: > Do you really need `-mavx512f`? What happens if you remove that (and make this a regular test), and remove the `defined(__AVX512F__)` code in the test? I think it will probably just work? That test only uses AVX512 via inline assembly which I don't think requires `-mavx512f`.

There's an interesting error that occurs when I remove the `-mavx512f` flag (and do not use `-march=native`): 

```
../src/test/x86/gdb_avx512.c:42:5: error: the register ‘zmm30’ cannot be clobbered in ‘asm’ for the current target
   42 |     asm volatile (
```

However, if I change the clobbered register to be something like `zmm9` and `zmm10` it _does_ compile without the flags, which means I can still test the contents of the registers is ok, but not all the way up to zmm31

theIDinside: It worked just fine to remove it, had to change what registers get clobbered though, but that's probably just me being unfamiliar with inline assembly, because I have no idea why that had to happen.

src/ExtraRegisters.cc
```diff
@@ -66,17 +67,91 @@ static bool reg_in_range(GdbServerRegister regno, GdbServerRegister low, GdbServ
...
+  // Feature bits given by CPUID
+  int8_t feature;
+  // Width in bytes in the xsave area
+  int8_t size;
+  // If the registers described by this descriptor is laid out in Hi16 ranges,
+  // value can be non-zero.
+  int8_t hi16_offset;
+  // Range of registers described by descriptor
+  GdbServerRegister base;
+  GdbServerRegister end_exclusive;
```
rocallahan, line 91: If you make it `end_inclusive` that simplifies the table definiition and the code is almost the same.

src/GdbServer.cc
```diff
@@ -186,28 +187,11 @@ static void maybe_singlestep_for_event(Task* t, GdbRequest* req) {
...
-      break;
-    case aarch64:
-      end = DREG_FPCR;
-      break;
-    default:
-      FATAL() << "Unknown architecture";
-      return;
-  }
+  const auto& registers = target_registers(regs.arch());
+  // TODO: Do we really need to allocate this every time for a reg request?q
```
rocallahan, line 191: Remove this. I think it's fine.

src/GdbServer.cc
```diff
@@ -2317,6 +2301,65 @@ void GdbServer::read_back_debugger_mem(DiversionSession& session) {
...
 }
 
+const vector<GdbServerRegister>& GdbServer::target_registers(
+    SupportedArch arch) {
+  if (!register_description.empty()) {
+    return register_description;
+  }
+
+  const auto add_range = [&](GdbServerRegister start, GdbServerRegister end) {
+    for (auto reg = start; reg < end; reg = GdbServerRegister(reg + 1)) {
```
rocallahan, line 2311: Make this `end_inclusive` and it will simplify the code.

src/test/x86/gdb_avx512.c
```diff
@@ -0,0 +1,71 @@
...
+        :
+        : "r"(0xff)
+        : "zmm7"
+    );
+#endif
+    return 0;
+}
+
+int
+main(void)
```
rocallahan, line 66: `int main(void)` in one line

src/test/x86/gdb_avx512.c
```diff
@@ -0,0 +1,71 @@
+#include <immintrin.h>
```
rocallahan, line 1: I think you can remove this.

rocallahan (changes requested):

theIDinside:
I'll rebase when we complete the review.

rocallahan:
I squashed it. Thanks!!!

sidkshatriya:
@theIDinside  
Hey I noticed something which I think probably should be remedied. 

In case you obtained (or obtained and then modified) the following xml files from [gdb](https://sourceware.org/gdb/current/)  sources:   
- `64bit-avx512.xml`
- `32bit-avx512.xml`
(these files were added to `rr` master in https://github.com/rr-debugger/rr/commit/2c1d0b237782116bfc7cd1fd8920685262cb1fa9 in this PR i.e. #3776 )

(see latest version for `64bit-avx512.xml` at https://sourceware.org/git/?p=binutils-gdb.git;a=blob_plain;f=gdb/features/i386/64bit-avx512.xml;hb=refs/heads/master )
(see latest version for `32bit-avx512.xml` here https://sourceware.org/git/?p=binutils-gdb.git;a=blob_plain;f=gdb/features/i386/32bit-avx512.xml;hb=refs/heads/master )

I'm noticing the Copyright and License header missing for the two XMLs in the `rr` git repository (compared to the `gdb` sources). Should this header (see below) be restored/added ( ( @rocallahan @khuey ) ? It would depend of course on how these XML files now in `rr` sources were obtained/generated. 

```
<!-- Copyright (C) 2014-2024 Free Software Foundation, Inc.

     Copying and distribution of this file, with or without modification,
     are permitted in any medium without royalty provided the copyright
     notice and this notice are preserved.  -->
```

khuey:
Yes if these were copied from the gdb repo they need the copyright headers.

theIDinside:
> Yes if these were copied from the gdb repo they need the copyright headers.

This had somehow slipped through my fingers. Addressed in https://github.com/rr-debugger/rr/pull/3934

PR #3822, opened 2024-09-12
check for EPERM in syscall perf_event_open

GitMensch:
EPERM is one of the explicit noted entries in man _syscall and is returned in docker environments without permissions (where /proc/sys/kernel is read-only)


grooverdan (approved):
good idea.

PR #3828, opened 2024-09-20
Don't assert on ticks overshoot during record on AMD

Keno:
On AMD, we regularly overshoot the programmed ticks counter. We print this as a warning earlier in the function. However, the warning is ineffective, because we have an assert with the same condition later in the function. Adjust the check for that error condition to only fire during replay, not record (where it is supposed to be non-fatal).


Keno:
Should resolve record failure seen at e.g. https://buildkite.com/julialang/julia-master/builds/40117#01920b75-60d9-42f0-a81c-c854b196ce20. cc @jeffbezanson

PR #3840, opened 2024-09-28
When using `PERF_RECORD_SWITCH`, allow ContextSwitchEvent initialization to fail

sidkshatriya:
The tracee may have died unexpectedly or been killed.

This commit allows the `cont_race` test to consistently pass when the test is run under all the following conditions:
(a) kernel.perf_event_paranoid = 2
(b) kernel supports PERF_RECORD_SWITCH (Linux kernel >= 6.10) 
(c) syscallbuf is enabled (so here we're talking only about `cont_race`
    and NOT `cont_race-no-syscallbuf` which doesn't need switch events anyways)

Without this commit, I noticed the occasional failure in (syscallbuf enabled) `cont_race`:
```
[FATAL src/PerfCounterBuffers.cc:32:allocate() errno: EBADF] Can't allocate memory for PT DATA area
```

### DIFF
--- a/src/ContextSwitchEvent.cc
+++ b/src/ContextSwitchEvent.cc
@@ -144,12 +144,15 @@ ContextSwitchEventStrategy ContextSwitchEvent::strategy() {
   return strat;
 }
 
-void ContextSwitchEvent::init(ScopedFd tracee_fd) {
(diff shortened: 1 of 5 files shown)


sidkshatriya:
(If you look at the code of the `cont_race` all the children are being killed by `do_killer()` which is running separately from `rr record`)

PR #3868, opened 2024-11-04
pack test: dont hunt for `mmap_pack_*` files in `latest-trace/` if the fs supports cloning

sidkshatriya:
`pack` test: dont hunt for `mmap_pack_*` files in `latest-trace/` if the fs supports cloning

Otherwise the `pack` test can incorrectly fail on, for example, btrfs:

  ```
  ls: cannot access 'latest-trace/mmap_pack_*': No such file or directory
  ```


sidkshatriya:
Rebased to the last master branch. CI is all green !

PR #3869, opened 2024-11-04
fix regression from #3867

GitMensch:
missing CXX string escape for \n


GitMensch:
... and this time the tests were run (and passed) on CI as well.

PR #3905, opened 2025-01-23
Repeat ARCH_XCOMP_GET_PERM if necessary as the relevant bit may be newly enabled.

khuey:
Ordinarily I'd just push something like this but it's not clear to me why the test is doing ARCH_GET_XCOMP_PERM before ARCH_REQ_XCOMP_PERM in the first place ...


rocallahan (commented):
I don't know, but this seems fine.

PR #3907, opened 2025-01-27
Add missing #include <sched.h>

rupprecht:
This is needed to verify the sched_param type.

The header is no longer transitively included via libc++ after https://github.com/llvm/llvm-project/commit/699f19605579f25083152a9ad21e14c2751d5d66


khuey:
Thanks!

PR #3911, opened 2025-01-31
Buffer epoll_ctl.

khuey:
I don't see any reason we can't do this, and it's worth a 6x improvement on a customer-provided testcase.


khuey:
Well, I accidentally pushed this, so if it's bad we'll have to back it out.

PR #3916, opened 2025-02-08
Make QListThreadsInStopReply adhere to thread id syntax

theIDinside:
In future work when RR supports multi process debug session/replays, this extension must send thread info in the correct format.

If the multi process extension is enabled all threads must contain p<pid.tid>. Debuggers that opt in to multiprocess features, must honor this.

### DIFF
--- a/src/GdbServerConnection.cc
+++ b/src/GdbServerConnection.cc
@@ -1755,6 +1755,47 @@ static int to_gdb_signum(int sig) {
   }
 }
 
+static constexpr void write_threads_extension(
+    bool multiprocess, stringstream& stream,
+    const vector<GdbServerConnection::ThreadInfo>& threads) {
+  stream << "threads:";
+  bool first = true;
+  if (multiprocess) {
+    for (const auto& thread : threads) {
+      if (!first) {
+        stream << ",";
+      }
+      first = false;
+      stream << "p" << thread.id.to_debugger_thread_id();
+    }
+  } else {
+    for (const auto& thread : threads) {
+      if (!first) {
+        stream << ",";
+      }
+      first = false;
+      stream << thread.id.tuid.tid();
+    }
+  }
+}
+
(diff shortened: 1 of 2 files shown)


theIDinside:
I wonder if this is correct, now that i think about it. It's difficult to reason about exactly what LLDB has had in mind with their extension. Don't pull this in. But I'm leaving it here, until I've investigated it further, if that's ok? Otherwise, close this PR @rocallahan and I'll resend it if/when I know more.

I can't explicitly find in the LLDB code base, where it parses the `threads:` list and if it uses the thread id syntax (mandated by the gdb multiprocess extension so it's... very icky that they arbitrarily format things the way they do).

theIDinside:
Now that I think about it, support for this extension is entirely unnecessary for RR. Even in a multi process scenario.

Why? Because it's only ever 1 thread that actually "stopped". In a scenario where a debugger is notified of the stop, due to the fact that this debugger knows it's interfacing with RR, it can safely query anything about the current thread list (of _all_ existing threads), and it can start reading things like data for unwinding the stack safely, because it *knows* that RR is stopped.

I will close this PR. Any debugger that will want to support "rr's multi process" support, can just ignore this LLDB extension entirely.

theIDinside:
Yeah, I can't really see why LLDB has chosen *not* to opt into the thread id syntax, when it supports the "multi process" flag, for this particular extension. But even so, this extension is entirely irrelevant for multiprocess debugging - in a multi process debugging environment, RR will inform the debugger of all threads, all processes that gets created and this will be *enforced* behind a `--multi` flag (or whatever we end up calling it), so that the debugger will actually know at all times and stops, what processes and threads it's dealing with (because it's always informed of creation and deaths).

In an even better scenario, when a debugger explicitly is written for interfacing with RR it could even query RR *before* even starting the session, since RR knows exactly when all this stuff happens.

PR #3956, opened 2025-05-21
use sys_gettid

GitMensch:
follow up to ec66b21 + 5d72fce fixing (part of) the build on RHEL8


GitMensch:
I have no clue why this fails - it does so in a test that I haven't adjusted...

PR #3957, opened 2025-05-21
fix build without PR_{GET,SET}_TAGGED_ADDR_CTRL defined

GitMensch:
follow up to 6ebf56d, note that both `PR_SET_TAGGED_ADDR_CTRL` and `PR_TAGGED_ADDR_ENABLE` are only available if `PR_GET_TAGGED_ADDR_CTRL` is defined


khuey:
It's better to define these missing constants in kernel_supplement.h than ifdef the functionality.

60ee7e04c50acc12b2a3f257454e76c3de2c35f5

PR #4003, opened 2025-08-13
Set dependency against only object

lissyx:
Not doing so may result in build failure:
```
$ sudo make signed
make -C /lib/modules/6.14.0-28-generic/build M= modules
make[1] : on entre dans le répertoire « /usr/src/linux-headers-6.14.0-28-generic »

make[3]: ***  Aucune règle pour fabriquer la cible « arch/x86/entry/syscalls/syscall_32.tbl », nécessaire pour « arch/x86/include/generated/uapi/asm/unistd_32.h ». Arrêt.
make[2]: *** [arch/x86/Makefile:277: archheaders] Error 2
make[1]: *** [Makefile:251: __sub-make] Error 2
make[1] : on quitte le répertoire « /usr/src/linux-headers-6.14.0-28-generic »
make: *** [Makefile:4: all] Error 2

### DIFF
--- a/third-party/zen-pmu-workaround/Makefile
+++ b/third-party/zen-pmu-workaround/Makefile
@@ -4,7 +4,7 @@ all:
 	make -C /lib/modules/$(shell uname -r)/build M=$(PWD) modules
 
 signed: $(obj-m)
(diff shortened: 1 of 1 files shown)


GitMensch:
There is no signed target in the Makefile (and never was in master), so I guess this PR can be closed.
_If_ the signed target is added then the dependency should be configured like here.

PR #4009, opened 2025-08-24
Alpine CI

GitMensch:
* only to be triggered manually (but to do so the file needs to be available on the main branch)
* ignores build errors for now, showing all issues instead of only one


GitMensch:
Just for reference the CI build from before removing LARGEFILE define: workflow run https://github.com/GitMensch/rr/actions/runs/17189977398/job/48764257227 also providing a [build log](https://github.com/GitMensch/rr/actions/runs/17189977398/artifacts/3838004691) of only the ninja part and the workflow run after https://github.com/GitMensch/rr/actions/runs/17210712395/job/48821516245 with its [ninja build log](https://github.com/GitMensch/rr/actions/runs/17210712395/artifacts/3843969557).

... and an updated one with the open referenced PRs included: https://github.com/GitMensch/rr/actions/runs/17749762907/artifacts/4018389855

GitMensch:
@rocallahan Would you merge this workflow definition?
If not: what is missing (other than a squash, but that can be done as part of the merge with a single click as well)?

Its sole purpose is to be run _manually_ from time to time (only the manual workflow-dispatch is enabled) - and any fork working on fixes would also be able to do that once pulled into master (the dispatch only works if the file is in master).

rocallahan:
Sure seems fine.

PR #4010, opened 2025-08-25
only define htonl where missing

GitMensch:
* glibc has that as a define
* muslc and others have that as a function, which leads to a build failure, as seen in the #4009 build log.
* Android API level 20 misses it


GitMensch:
I claim that the failing test

>  	966 - alternate_thread_diversion-no-syscallbuf (Failed)

is totally unrelated as this PR changes a single different test.

PR #4020, opened 2025-09-15
provide constants for ioctl in muslc

GitMensch:
see https://man.archlinux.org/man/core/man-pages/TCSETS.2const.en


GitMensch:
Those are just not defined... so I'll #ifndef them away in a new PR.

PR #4035, opened 2025-12-19
Fix resolve_address to account for multiple mappings of the same file…

pcc:
… offset

Some linkers such as lld will create program headers with multiple mappings of the same file offset. This can lead to problems when a symbol of interest to rr, such as __aarch64_ldadd4_relax, is covered by more than one mapping, as that will lead to us finding the function in multiple mappings. For that symbol in particular, we can end up misinterpreting the instructions in the wrong mapping and incorrectly computing an address to write to, which can lead to an assertion failure, or worse, silent memory corruption. Fix it by changing resolve_address to check whether the mapping is the correct one (fully covers the appropriate program header and has the same memory permissions) before returning the address.

### DIFF
--- a/src/ElfReader.cc
+++ b/src/ElfReader.cc
@@ -28,7 +28,7 @@ public:
   virtual Debugaltlink read_debugaltlink() = 0;
   virtual string read_buildid() = 0;
   virtual string read_interp() = 0;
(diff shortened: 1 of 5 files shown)


rocallahan:
Thanks!

PR #4052, opened 2026-02-13
rename crc32 to rr_crc32 to avoid conflict

benquike:


rocallahan:
I just did this already :-)

PR #4066, opened 2026-04-14
Support preadv2 and pwritev2 syscalls.

### LINKED ISSUE #3193: Support `preadv2/pwritev2`
These were added in Kernel 4.6 and they are quite similar to preadv and pwritev but adds a flags argument to tweak the behaviour.

The newest flag is `RWF_APPEND` which was added in 4.16.

I'm not sure exactly how these would be implemented but I noticed the application I was attempting to debug used these. So it would be nice if it was supported.

azat:
These were marked as UnsupportedSyscall, causing rr to expect ENOSYS. On modern kernels that support them (Linux 4.6+), the syscall succeeds and rr hits a FATAL assertion. Handle them the same way as their v1 counterparts (preadv/pwritev).

Fixes #3193


src/record_syscall.cc
```diff
@@ -4556,7 +4557,10 @@ static Switchable rec_prepare_syscall_arch(RecordTask* t,
     case Arch::readv:
     /* ssize_t preadv(int fd, const struct iovec *iov, int iovcnt,
                       off_t offset); */
-    case Arch::preadv: {
+    case Arch::preadv:
+    /* ssize_t preadv2(int fd, const struct iovec *iov, int iovcnt,
+                       off_t offset, int flags); */
+    case Arch::preadv2: {
```
rocallahan, line 4563: It's possible that some future `RWF_` flag will introduce behaviour that breaks rr. So it would be a good idea to define a set of flags that we know are ok with rr with `preadv2` (all the flags currently defined, I think), and fail here with `EINVAL` if some flag outside that set is passed in. See below for why this is especially important for `pwritev2`.

src/Task.cc
```diff
@@ -761,6 +761,7 @@ void Task::on_syscall_exit_arch(int syscallno, const Registers& regs) {
     }
 
     case Arch::pwritev:
+    case Arch::pwritev2:
```
rocallahan, line 764: If the `RWF_APPEND` flag is set then the kernel ignores the user's offset, but we'll compute an incorrect offset below and things will go wrong. (I think we already have an existing bug if the user passes `O_APPEND` to `open()`; in that case `pwritev` will ignore the offset and we compute the wrong offset.)

So let's do what I said above and specify a mask of flags we know we can handle correctly --- I think that's everything currently defined except for `RWF_APPEND` --- and return `EINVAL` if any other flags are set. That will require adding code in `record_syscall.cc` for `pwritev2`.

azat: Thanks for taking a look! Fixed

rocallahan:
Thanks!!!

PR #4067, opened 2026-04-15
Pass through and record PR_SVE system calls on arm

earthling-amzn:
Hotspot's JIT compilers use these system calls on ARMv8 to configure and detect SVE registers. These are not currently supported by `rr`. We could workaround this in the JVM by setting `-XX:UseSVE=0`, but it would also be nice to have `rr` support this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1450,6 +1450,7 @@ set(BASIC_TESTS
   sysinfo
   syslog
   arm/tagged_addr_ctrl
+  arm/sve_vl
   tgkill
   thread_yield
   timer
--- a/src/kernel_supplement.h
+++ b/src/kernel_supplement.h
@@ -520,6 +520,17 @@ enum {
 #define BLKGETNRZONES _IOR(0x12, 133, __u32)
 #endif
 
+// New in the 4.15 kernel
+#ifndef PR_SVE_SET_VL
+#define PR_SVE_SET_VL 50
+#endif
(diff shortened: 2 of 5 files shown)


khuey:
Out of curiosity and my own lack of familiarity with SVE, is the "correct" value of PR_SVE_GET_VL easily observable by executing some SVE instruction(s)?

earthling-amzn:
Here's a patch for the test case in this PR that, I think, shows what you are asking:
```
diff --git a/src/test/arm/sve_vl.c b/src/test/arm/sve_vl.c
index 447dabe4..9056c4a3 100644
--- a/src/test/arm/sve_vl.c
+++ b/src/test/arm/sve_vl.c
@@ -35,6 +35,12 @@ int main(void) {
   int vl2 = prctl(PR_SVE_GET_VL);
   test_assert((vl2 & PR_SVE_VL_LEN_MASK) == len);
 
+  /* Verify the prctl value matches the hardware vector length.
+     RDVL returns the SVE vector length in bytes for the current thread. */
+  long rdvl;
+  __asm__ volatile(".arch armv8-a+sve\n\trdvl %0, #1" : "=r"(rdvl));
+  test_assert(rdvl == len);
+
   atomic_puts("EXIT-SUCCESS");
   return 0;
 }
```
It adds an assertion that the value returned from `prctl(PR_SVE_GET_VL)` matches what the instruction to read vector length `rdvl` returns. On my host (graviton-3), the vector length is 32 bytes (256 bits).

PR #4073, opened 2026-05-13
Advertise pointer authentication support to GDB

pcc:
GDB uses the values of the provided pointer authentication mask registers to mask return addresses from the stack trace. Because we weren't providing these masks, the return addresses in the stack trace were incorrect in programs built with -mbranch-protection=pac-ret, such as most of Fedora. Fix it by advertising pointer authentication to GDB, and providing the correct masks.

Fixes the following tests on PAC platforms:

fork_exec_info_thr
fork_exec_info_thr-no-syscallbuf
vdso_clock_gettime_stack
vdso_time_stack

### DIFF
--- a/src/ExtraRegisters.cc
+++ b/src/ExtraRegisters.cc
@@ -275,6 +275,21 @@ size_t ExtraRegisters::read_register(uint8_t* buf, GdbServerRegister regno,
     } else if (regno == DREG_FPCR) {
       reg_data = RegData(offsetof(ARM64Arch::user_fpsimd_state, fpcr),
                          sizeof(uint32_t));
+#ifdef __aarch64__
+    } else if (regno == DREG_PAUTH_DMASK || regno == DREG_PAUTH_CMASK) {
+      uint64_t ptr = 1ULL << 55;
+      // The XPAC instruction will copy bit 55 of the argument into the PAC mask
+      // bits, so ptr will be set to the mask plus bit 55.
+      if (regno == DREG_PAUTH_DMASK) {
+        __asm__ __volatile__("xpacd %0" : "+r"(ptr));
+      } else {
+        __asm__ __volatile__("xpaci %0" : "+r"(ptr));
+      }
+      uint64_t mask = ptr & ~(1ULL << 55);
+      *defined = true;
+      memcpy(buf, &mask, sizeof(mask));
+      return sizeof(mask);
+#endif
     } else {
       *defined = false;
       return 0;
--- a/src/GdbServerConnection.cc
+++ b/src/GdbServerConnection.cc
@@ -18,6 +18,7 @@
 #include <poll.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/auxv.h>
 #include <sys/mman.h>
 #include <sys/socket.h>
 #include <sys/types.h>
@@ -32,6 +33,7 @@
 #include "ScopedFd.h"
 #include "TargetDescription.h"
 #include "core.h"
+#include "kernel_supplement.h"
 #include "log.h"
 
 using namespace std;
@@ -106,6 +108,11 @@ static uint32_t get_cpu_features(SupportedArch arch) {
     }
     case aarch64:
       cpu_features = GdbServerConnection::CPU_AARCH64;
+#ifdef __aarch64__
+      if (getauxval(AT_HWCAP) & HWCAP_PACA) {
+        cpu_features |= GdbServerConnection::CPU_PAUTH;
+      }
+#endif
       break;
     default:
       FATAL() << "Unknown architecture";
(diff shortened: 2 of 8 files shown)


khuey:
~~I think you forgot to git add pauth.xml.~~

Apparently that's already in the tree.

pcc:
It's part of gdb
```
$ find | grep pauth.xml
./third-party/gdb/aarch64-pauth.xml
```

src/ExtraRegisters.cc
```diff
@@ -275,6 +275,21 @@ size_t ExtraRegisters::read_register(uint8_t* buf, GdbServerRegister regno,
     } else if (regno == DREG_FPCR) {
       reg_data = RegData(offsetof(ARM64Arch::user_fpsimd_state, fpcr),
                          sizeof(uint32_t));
+#ifdef __aarch64__
+    } else if (regno == DREG_PAUTH_DMASK || regno == DREG_PAUTH_CMASK) {
+      uint64_t ptr = 1ULL << 55;
+      // The XPAC instruction will copy bit 55 of the argument into the PAC mask
+      // bits, so ptr will be set to the mask plus bit 55.
+      if (regno == DREG_PAUTH_DMASK) {
+        __asm__ __volatile__(".arch armv8.3-a\nxpacd %0" : "+r"(ptr));
```
khuey: We require armv8.3-a anyways so do we need the arch directives?

src/GdbServerConnection.cc
```diff
@@ -18,11 +18,18 @@
...
 #include <stdlib.h>
+#include <sys/auxv.h>
 #include <sys/mman.h>
 #include <sys/socket.h>
 #include <sys/types.h>
 #include <unistd.h>
 
+#ifdef __aarch64__
+#ifndef HWCAP_PACA
+#define HWCAP_PACA (1 << 30)
```
khuey: Nitpicking here but put this in kernel_supplement.h please.

src/Registers.cc
```diff
@@ -180,6 +180,10 @@ RegisterInfo<rr::ARM64Arch>::Table RegisterInfo<rr::ARM64Arch>::registers = {
   // comparison errors. This seems to be a unused/undocumented bit in SPSR as per aarch64
   // documentation anyways so ignore it.
   RV_AARCH64_WITH_MASK(CPSR, pstate, 0xffffffffLL & ~AARCH64_DBG_SPSR_SS & ~AARCH64_DBG_SPSR_11, 4),
+  // PAuth masks. These aren't real registers, so we set the offset/size to 0
+  // and use custom handling in read_register().
+  RegisterInit(DREG_PAUTH_DMASK, RegisterValue("pauth_dmask", 0, 0)),
+  RegisterInit(DREG_PAUTH_CMASK, RegisterValue("pauth_cmask", 0, 0)),
```
khuey: I don't understand why this is needed. We don't need this for e.g. DREG_FPCR.

src/ExtraRegisters.cc
pcc: No, removed.

src/GdbServerConnection.cc
pcc: Done

src/Registers.cc
pcc: Yeah, not necessary, removed. I got confused and thought that `ExtraRegisters` was a subset of `Registers` when it's actually either/or.

We may consider removing the `Registers` code path and handling everything through what is now `ExtraRegisters`, which would be renamed to `Registers`. It would be a bit more code for each currently non-extra register but it would make it clearer how each register is handled.

pcc:
Looks like CI is failing (some of the tests SIGABRT) but I don't have any failures like that on my end. Would it be possible for someone with access to the CI to take a look?

rocallahan:
> Would it be possible for someone with access to the CI to take a look?

I can do that, and I will try to do that soon. But it is relatively easy to replicate the CI environment. We use AWS `c6g.8xlarge` instances and run setup scripts `scripts/github-actions-build.sh` and `scripts/github-actions-test.sh`. Ubuntu 24 LTS.

khuey:
```
Program received signal SIGILL, Illegal instruction.
0x0000aaaaaae7e6e8 in rr::ExtraRegisters::read_register (this=0xaaaaab3b1a90, buf=0xffffffffc968 "",
    regno=rr::DREG_64_YMM8H, defined=0xffffffffc990) at /home/ubuntu/rr/src/ExtraRegisters.cc:284
284             __asm__ __volatile__("xpacd %0" : "+r"(ptr));
```

CI runs on Graviton2 and it looks like pointer authentication was added in Graviton3.

khuey:
Ignore my previous comment about HWCAP2. That's what I get for not double checking Gemini.

The cset I just pushed should fix the CI issues.

khuey:
Thanks for the PR!

pcc:
> CI runs on Graviton2 and it looks like pointer authentication was added in Graviton3.

So does it mean that we shouldn't be building with `-march=armv8.3-a`?

khuey:
@rocallahan We should probably bump the runners to Graviton3 instances.

rocallahan:
> @rocallahan We should probably bump the runners to Graviton3 instances.

Maybe we should have both?

khuey:
> > @rocallahan We should probably bump the runners to Graviton3 instances.
> 
> Maybe we should have both?

Well we're building with -march=armv8.3-a but running on a processor that only supports the v8.2-A ISA.

rocallahan:
Ok I updated the runner function to spawn `m7g.8xlarge` instances.

Issue #4081, opened 2026-06-24
performance issue in print_reg_value() when using rr to trace function calls


core-explorer:
I want to get a heaptrack-compatible heap profile out of an rr recording (the recording was done without heaptrack).

To do that, I have a driver that speaks the GDB remote protocol directly to rr, sets breakpoints on malloc, free, realloc and related functions
and basically does continue/get_registers in a tight loop.

This is quite a different usage profile from a human sitting in front of GDB and typing in commands. I build rr in a Release configuration, with optimizations enabled.

When rr encodes register values as binary strings, it calls `snprintf` once per byte per register.

https://github.com/rr-debugger/rr/blob/035e5e8d996971a84c9a0094f6701f368a35d7c2/src/GdbServerConnection.cc#L628-L630

When I replaced that call with a table-lookup per nibble the total wall clock time for my program using rr dropped by 50%.

This is the kind of optimization that leads to duplicated lookup-tables everywhere (and forgotten zero-terminators), so there should probably be some write_as_hex_string functions that can be re-used in Registers.cc and GdbServerConnection.cc

rocallahan:
Sure, we'd take a PR for this.

core-explorer:
solved by pull request #4082, thanks

PR #4082, opened 2026-06-25
GdbServerConnection.cc: do not call sprintf in a loop

core-explorer:
### Add a function `write_hex_string` that writes the input's hex representation into a buffer.

The motivation for this is the performance impact of `print_reg_value` when repeatedly getting registers on a system with AVX-512 (issue #4081).

Other functions that call `sprintf` or `snprintf` in a loop have been changed, the motivation for that is internal consistence, not performance.

In the functions `string_to_hex` and `to_string`, the size of the buffer backing the string is specified as `length() + 1`, which may look suspect for a `std::string`. This does not invoke undefined behavior in C++17, which blessed overwriting the zero terminator with `'\0'`, that is a rather obscure corner of C++ (https://cplusplus.github.io/LWG/lwg-defects.html#2475)

### DIFF
--- a/src/GdbServerConnection.cc
+++ b/src/GdbServerConnection.cc
@@ -275,14 +275,35 @@ void GdbServerConnection::write_binary_packet(const char* pfx, const uint8_t* da
   return write_packet_bytes(buf.data(), buf_num_bytes);
 }
 
-static string string_to_hex(const string& s) {
-  stringstream sstr;
-  for (char ch : s) {
-    char buf[16];
-    sprintf(buf, "%02x", ch);
-    sstr << buf;
+static void parser_assert(bool cond) {
+  if (!cond) {
+    fputs("Failed to parse debugger request\n", stderr);
+    DEBUG_ASSERT(false);
+    exit(2);
   }
-  return sstr.str();
+}
+
+/**
+ * For the first |input_len| bytes in |input|, append two characters to |buf| as if
+ * snprintf("%02x") would, then write a zero terminator.
+ */
+static void write_hex_string(const uint8_t* input, size_t input_len,
+                             char* buf, size_t buf_len) {
+  static constexpr char hex_string_table[] = "0123456789abcdef";
+  parser_assert(buf_len >= 2 * input_len + 1);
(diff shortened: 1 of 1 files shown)


src/GdbServerConnection.cc
```diff
@@ -275,14 +275,34 @@ void GdbServerConnection::write_binary_packet(const char* pfx, const uint8_t* da
...
+    buf[2 * i + 0] = hex_string_table[input[i] >> 4];
+    buf[2 * i + 1] = hex_string_table[input[i] & 0xf];
   }
-  return sstr.str();
+  buf[2 * input_len] = '\0';
+}
+
+static string string_to_hex(const string& s) {
+  string output;
+  output.resize(2 * s.size());
```
rocallahan: It's a little weird to overwrite std::string's trailing null byte, so I'd prefer to make this `2 * s.size() + 1`.

src/GdbServerConnection.cc
```diff
@@ -1322,17 +1329,14 @@ bool GdbServerConnection::process_vpacket(char* payload) {
...
-      break;
-    }
-    char buf[3];
-    sprintf(buf, "%02x", bytes[i]);
-    ss << buf;
-  }
-  return ss.str();
+  string output;
+  size_t num_bytes = std::min(max_len, bytes.size());
+  output.resize(2 * num_bytes);
```
rocallahan: Same here.

rocallahan (changes requested):

src/GdbServerConnection.cc
core-explorer: I agree it's weird. I'm in favor of avoid weird-looking code where avoidable.
In this case, it would have to be
```
resize(2*n+1); 
write_hex_string(); 
resize(2*n);
```
as otherwise we get a string with `size() != strlen(c_str())`.

An alternative is to not write the trailing zero in `write_hex_string` and have that the caller's responsibility.

src/GdbServerConnection.cc
core-explorer: I changed both instances to resize twice

PR #4083, opened 2026-07-03
Add informational commands: absolute-time, when-end, info recording

timo:
absolute-time gives the absolute monotonic clock time from the current event.

This is useful when you want to align events with a perf recording that also uses the monotonic clock (you still have to scale by 1000000000 of course.)

when-end gives the event number of the last event in the whole recording.

This is useful when running a replay with non-stopping breakpoints and displaying the "progress" of the replay to the user.

info recording prints out the path of the recording folder.

This is useful for running additional `rr replay` sessions from an existing one, or for calling `rr dump` to get data that is not otherwise surfaced by any gdb commands.

The next thing I am going to do is write tests for these commands, but I'd like to already PR the changes to get feedback on whether the implementation and the names of the commands are OK.

Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1605,6 +1605,7 @@ set(TESTS_WITH_PROGRAM
   ignored_sigsegv
   ignore_nested
   immediate_restart
+  info_recording
   x86/int3_ok
   interrupt
   intr_ptrace_decline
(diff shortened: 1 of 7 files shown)


timo:
Tests written, I think the code is ready for a review now.

rocallahan:
Sorry about the delay.
Please split this into 3 PRs, one for each command. And in each PR, squash all the changes together into a single commit. Thanks!

timo:
obsoleted by https://github.com/rr-debugger/rr/pull/4084 https://github.com/rr-debugger/rr/pull/4085 and https://github.com/rr-debugger/rr/pull/4086

PR #4084, opened 2026-07-20
Add "when-end" command to get the number of the last event

timo:
when-end gives the event number of the last event in the whole recording.

This is useful when running a replay with non-stopping breakpoints and displaying the "progress" of the replay to the user.

I'm not actually sure if `result.type() == TraceTaskEvent::Type::NONE` is an idiomatic way to see if the `read_task_event` call returned the "default value" of that type. But it does at least seem to work in the cases I've tested.


rocallahan:
I think this caused that test failure in `when.py` ... on Aarch64 ... with LLDB :-(.
I guess you could look into it by pushing test PRs with extra logging, maybe.

PR #4085, opened 2026-07-20
"info recording" command that gives the path to the recording

timo:
info recording prints out the path of the recording folder.

This is useful for running additional rr replay sessions from an existing one, or for calling rr dump to get data that is not otherwise surfaced by any gdb commands.

### DIFF
--- a/src/DebuggerExtensionCommand.cc
+++ b/src/DebuggerExtensionCommand.cc
@@ -4,6 +4,7 @@
 
 #include "ReplayTask.h"
 #include "log.h"
+#include "util.h"
 
 using namespace std;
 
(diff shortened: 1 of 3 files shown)


src/test/info_recording.c
rocallahan, line 1: I don't think you need your own test program here. You can use `simple.c` like `when.run` does.

src/DebuggerExtensionCommand.cc
```diff
@@ -56,6 +57,21 @@ static SimpleDebuggerExtensionCommand when_tid(
...
     });
 
+static SimpleDebuggerExtensionCommand info_recording(
+    "info recording",
+    "Print the path of the recording RR is"
+    " currently replaying.",
+    [](GdbServer&, Task* t, const vector<string>&) {
+      if (!t->session().is_replaying()) {
+        return DebuggerExtensionCommandHandler::cmd_end_diversion();
+      }
```
rocallahan: I don't think you need to end the diversion for this.

rocallahan (changes requested):
Looks good modulo these minor comments.

src/test/info_recording.c
timo: done

src/DebuggerExtensionCommand.cc
timo: removed.

rocallahan: Looks to me like it's still here.

rocallahan (changes requested):

src/DebuggerExtensionCommand.cc
timo: whoops?! not sure how I got confused there. now I've double-checked and it really is gone!

GitMensch:
Just wondering - Wouldn't it be useful to also output more data - like the amount of events (the when-end?), the current event (= recording equivalence to "program stopped at ..." from "info program"), maybe the recording size on disk, ...?

PR #4086, opened 2026-07-20
Add command "absolute-time" to get value of monotonic clock

timo:
absolute-time gives the absolute monotonic clock time from the current event.

This is useful when you want to align events with a perf recording that also uses the monotonic clock (you still have to scale by 1000000000 of course.)

### DIFF
--- a/src/DebuggerExtensionCommand.cc
+++ b/src/DebuggerExtensionCommand.cc
@@ -3,6 +3,7 @@
 #include "DebuggerExtensionCommand.h"
 
 #include "ReplayTask.h"
+#include "TraceFrame.h"
 #include "log.h"
 
 using namespace std;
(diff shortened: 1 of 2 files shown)


src/DebuggerExtensionCommand.cc
```diff
@@ -25,6 +26,21 @@ static SimpleDebuggerExtensionCommand elapsed_time(
...
     });
 
+static SimpleDebuggerExtensionCommand absolute_time(
+    "absolute-time",
+    "Print absolute time (in seconds) from the monotonic clock in the"
+    " 'record' timeline.",
+    [](GdbServer&, Task* t, const vector<string>&) {
+      if (!t->session().is_replaying()) {
+        return DebuggerExtensionCommandHandler::cmd_end_diversion();
+      }
```
rocallahan: I don't think you need to end the diversion here.

rocallahan (changes requested):
Looks good other than this comment.

src/DebuggerExtensionCommand.cc
timo: removed it

PR #4087, opened 2026-07-21
Update CTestCostData.txt.

bernhardu:


bernhardu:
This updates the cost data truncated to full seconds, combining x86_64 and arm64, using the longer time.

bernhardu:
While thinking about it, would it be sufficient to only store those lines that last longer than e.g. 10 seconds?
That would only give a file with 102 lines.

PR #4088, opened 2026-07-30
Add support for Qualcomm Oryon aarch64 CPU

takikawa:
This PR adds support for Qualcomm Oryon arm64 chips, AKA Snapdragon X1, that are used in various laptops like the Thinkpad T14s.

With this patch, the `counters-test` test succeeds for me and most of the main tests also pass on my machine (with a Snapdragon X1E-78-100). 5 tests do fail which I looked into:
  * `tls`: these 2 seem to be related to an existing gdb bug (``./gdb/regcache.c:810: internal-error: cooked_read: Assertion `regnum >= 0 && regnum < m_descr->nr_cooked_registers' failed.``)
  * `fork_exec_info_thr` is backtrace related (`Backtrace stopped: Cannot access memory at address 0x681ffc10`)
  * `record_replay`: these 2 may be related to pointer authentication (SIGILL due to ptrace with `PTRACE_SETREGSET` of `NT_ARM_PACA_KEYS`)

In any case I don't think they're caused by this patch itself.

Reference for the part number: https://github.com/torvalds/linux/blob/master/arch/arm64/include/asm/cputype.h#L132


rocallahan:
Thanks!

Issue #4072, opened 2026-05-07
rr should open all PMUs


pcc:
https://github.com/google/benchmark/pull/2123

rr is affected by the same issue and should open all PMUs (or at least the PMU for the pinned CPU).

Symptom:
```
[FATAL src/PerfCounters.cc:519:check_working_counters() errno: EDOM] 
Got 0 branch events, expected at least 500.

The hardware performance counter seems to not be working on cpu 13
Check that hardware performance counters are working by running
  perf stat -e apple_blizzard_pmu/r90/ true
and checking that it reports a nonzero number of events.
If performance counters seem to be working with 'perf', file an
rr issue, otherwise check your hardware/OS/VM configuration. Also
check that other software is not using performance counters on
this CPU.
=== Start rr backtrace:
rr() [0x4bb430]
rr() [0x4bb7ac]
rr(_ZN2rr12PerfCounters5startEPNS_4TaskEl+0x80c) [0x4e8b98]
rr(_ZN2rr4Task16resume_executionENS_13ResumeRequestENS_11WaitRequestENS_12TicksRequestEi+0x5ac) [0x5b91ac]
rr(_ZN2rr13RecordSession11record_stepEv+0x3b4) [0x50d120]
rr(_ZN2rr13RecordCommand3runERSt6vectorINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESaIS7_EE+0xd44) [0x4fbbf0]
rr(main+0x170) [0x431d10]
/lib64/libc.so.6(+0x2621c) [0xffff6107621c]
/lib64/libc.so.6(__libc_start_main+0x9c) [0xffff610762fc]
rr(_start+0x30) [0x433070]
=== End rr backtrace
```

khuey:
Is this specific to Apple's PMUs or is this a general Linux thing?

pcc:
My understanding is that this can happen on any system where there are multiple CPU PMUs, not just Apple's. It might even be possible on x86 as some systems have started to implement heterogeneous cores. @captain5050 to confirm.

yuyichao:
The intention of the code was to open the PMU for the core the thread is pinned to. Maybe that breaks?

Otoh, I feel like the PMU for the E cores simply stops working anymore and I’m 80% I’ve verified it with perf directly…

yuyichao:
[...]

PR #4091, opened 2026-08-10
add entries for ARM Cortex-A720 + A750 CPUs

pH14:
This PR adds support for Cortex-A720 cores, marks the Cortex-A520 as unsupported, and possibly fixes the regression noted in https://github.com/rr-debugger/rr/issues/4072

---

I tested out `rr` on an MS-R1 box which has a CIX P1 chip, which has both Cortex-A720 and Cortex-A520 cores. The Cortex-A720 PMU passed all tests, and the Cortex-A520 PMU intermittently overcounted branches so it's marked unsupported.

While doing the validation, I found two additional places that needed small tweaks:

1. Preserving named PMU event types -- this change in `PerfCounters.cc` _I think_  fixes the regression noted in https://github.com/rr-debugger/rr/issues/4072 where AArch64 systems with heterogeneous cores could result in seeing no branch events

2. Update automatic CPU selection -- the existing logic could result in a A520 core being selected. I've updated the selection code to first filter for only supported CPUs (and then filter to P cores if available as before).

---

I've read the contributor guidelines, but admittedly this is my first commit here. Happy to make any changes necessary, and thanks for the great project 🙇


rocallahan:
Looks great, thanks!!

PR #4092, opened 2026-08-14
Introduce `rr dump --json`, dump trace in JSONL, one object per line

timo:
Combines with some of the other flags:
* `--syscallbuf` adds a `syscalls` key
* `--mmaps` adds a `maps` key
* `--task_events` adds a `task_events` key`
* `--recorded-metadata` adds a `metadata` key

Furthermore, also passing `--raw` will add extra output to some of the above:
* `syscallbuf_flush_raw` key with the header of a syscallbuf flush
* `raw` key with the syscall data for each syscall
* `raw` key with mmap record data to mmaps entries

On top of all that, a `--no-regs` flag is introduced that skips outputting
register values in the regular, the raw, or the json output.

This also moves write_hex_string from being a static function inside
GdbServerConnection to live in src/util.cc instead.

Here's some example output:

https://gist.github.com/timo/d6dd3b2762d78a9ee468b3cb554fa6b9

Open questions:

* Currently, some functions are duplicated as a _json variant, some functions check `flags.json_dump` in-line. Is there a preference?
* The current dump function ignores when `frame.event().record_extra_regs()` is true; do we care about that?
* The initial thing that caused me (well, really dzaima) to look into the code in the first place was that the header of `--raw` still has three entries in it that correspond to field that are not being output any more. Is that for a separate pull request, or fine to include here?
* The help for the command says that `--raw` outputs in a different format, and `--json` makes the same claim, so it's not obvious that they can be combined. The current behaviour of `--raw` and `--json` together may want to get a different flag and then `--raw` could conflict with `--json`
  * there could instead be a `--format=raw`/`--format=json` option instead? though a short flag for json would be nice

### DIFF
--- a/src/DumpCommand.cc
+++ b/src/DumpCommand.cc
@@ -10,6 +10,7 @@
 #include <limits>
 #include <unordered_map>
 
+#include "Event.h"
 #include "preload/preload_interface.h"
 
 #include "AddressSpace.h"
@@ -24,6 +25,27 @@
 using namespace std;
 
 namespace rr {
+
+struct JsonArrayAtKeyWriter {
+  public:
+  bool key_written = false;
+  bool write_comma = false;
+
+  void ensure_key(FILE *out, const char *key) {
+    if (!key_written) {
+      key_written = true;
+      fprintf(out, ", \"%s\": [", key);
+    }
+  };
+
+  void maybe_comma(FILE *out) {
+    if (!write_comma) {
+      write_comma = true;
+    } else {
+      fprintf(out, ", ");
+    }
+  };
+};
 
 class DumpCommand : public Command {
 public:
@@ -45,9 +67,12 @@ DumpCommand DumpCommand::singleton(
     "  -e, --task-events          dump task events\n"
     "  -m, --recorded-metadata    dump recorded data metadata\n"
     "  -p, --mmaps                dump mmap data\n"
+    "  -R, --no-regs              skip dumping register values\n"
     "  -r, --raw                  dump trace frames in a more easily\n"
     "                             machine-parseable format instead of the\n"
     "                             default human-readable format\n"
+    "  -j, --json                 dump trace frames in jsonl, i.e. one\n"
+    "                             json object per line\n"
     "  -s, --statistics           dump statistics about the trace\n"
     "  -t, --tid=<pid>            dump events only for the specified tid\n");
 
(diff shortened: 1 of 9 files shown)


src/DumpCommand.cc
```diff
@@ -25,6 +26,27 @@ using namespace std;
 
 namespace rr {
 
+struct JsonArrayAtKeyWriter {
+  public:
```
rocallahan, line 30: Don't need `public:`, it's the default for struct

src/DumpCommand.cc
```diff
@@ -25,6 +26,27 @@ using namespace std;
 
 namespace rr {
 
+struct JsonArrayAtKeyWriter {
```
rocallahan, line 29: Document what this struct is for and what its parts mean

src/DumpCommand.cc
```diff
@@ -122,47 +155,84 @@ static void dump_syscallbuf_data(TraceReader& trace, FILE* out,
...
+              syscall_name(record->syscallno, frame.regs().arch()).c_str(),
+              (long)record->ret, (long)record->size,
+              record->desched ? ", desched:1" : "",
+              record->replay_assist ? ", replay_assist:1" : "");
+    }
     if (flags.raw_dump) {
-      fprintf(out, "  ");
-      for (unsigned long i = 0; i < record->size; ++i) {
-        fprintf(out, "%2.2x", *(record_ptr + (uintptr_t)i));
+      auto record_out = std::vector<char>();
```
rocallahan, line 191: make this `std::string`.

src/DumpCommand.cc
```diff
@@ -25,6 +26,27 @@ using namespace std;
 
 namespace rr {
 
+struct JsonArrayAtKeyWriter {
+  public:
+  bool key_written = false;
+  bool write_comma = false;
+
+  void ensure_key(FILE *out, const char *key) {
```
rocallahan, line 34: This is weird. Why would the caller not know if a key has been emitted yet?

rocallahan (commented):
I think this is basically fine as-is, just some minor comments. Sorry for the delay, I've been traveling.

rocallahan:
Sorry for the delay, I was traveling.

> * Currently, some functions are duplicated as a _json variant, some functions check `flags.json_dump` in-line. Is there a preference?

No, I think having a mix like you have is OK.

> * The current dump function ignores when `frame.event().record_extra_regs()` is true; do we care about that?

No, at least, definitely not in this PR.

> * The initial thing that caused me (well, really dzaima) to look into the code in the first place was that the header of `--raw` still has three entries in it that correspond to field that are not being output any more. Is that for a separate pull request, or fine to include here?

Separate PR please.

> * The help for the command says that `--raw` outputs in a different format, and `--json` makes the same claim, so it's not obvious that they can be combined. The current behaviour of `--raw` and `--json` together may want to get a different flag and then `--raw` could conflict with `--json`

Yeah I think it doesn't really make sense to combine them and we shouldn't support that.

>   * there could instead be a `--format=raw`/`--format=json` option instead? though a short flag for json would be nice

I'm OK with having one flag per format.

